PR #1, opened 2023-01-29
Add Visual Studio project and solution generator to scons as a build option

Spartan322:
Called via `scons vsproj=yes`. Also updated README.md in reference to it.

This requires validation that the generator properly works on Windows with Microsoft's Visual Studio installed.


BrickPi:
Has there been any update to either scons or GDC++ documentation regarding building VS projekts?

Spartan322:
> Has there been any update to either scons or GDC++ documentation regarding building VS projekts?

Not that I've seen, GDExtensions and godot-cpp itself will require some TLC from MSVC folks, hopefully this should happen just like it did with the engine development itself but right now that kinda isn't the case.

Spartan322:
Closed as I can be expected to manage this, needs someone who knows Scons' vsproj builder better then me, I can't even test it without Windows which I don't use.

PR #4, opened 2023-02-05
Fix up cmake

BrickPi:
Worked on the CMakeLists.txt, updated the README, removed ref to Scons, and temporarily removed the Scons workflow until it gets updated for cmake.


Spartan322:
Relevant PRs:
godotengine/godot-cpp#1041
godotengine/godot-cpp#957
godotengine/godot-cpp#344

PR #6, opened 2023-02-09
Update Godot 4 to Release Candidate 1

Spartan322:

PR #7, opened 2023-02-10
Implements a functioning user settings UI

Spartan322:
Relies on: #6 

## Implements

- Resolution selector
- Screen Mode selector
- Monitor Selection selector
- Master Volume slider
- Music Volume slider
- SFX Volume slider

Note: These all have an effect on the application.

Handles user saving and loading user data according to all these settings, each also has default values and can be reset. Settings will also be saved if the options menu is open and the application is closed or if Godot's crash handler is enabled and the program crashes.

Should fulfill: SS-58, SS-61, SS-6, SS-9, SS-10, SS-11, SS-13, UI-11, UI-12, UI-19, UI-44, UI-47, UI-22


Hop311 (approved):
LGTM, the functionality is working for the six specified settings and the settings state persists through closing and reopening the application.

PR #8, opened 2023-02-12
Exploration of uniting Godot UI with Stateful C++ Objects via GDExtension

ClarkeCode:


extension/src/TestSingleton.cpp
```diff
@@ -6,31 +6,26 @@
 using namespace godot;
 using namespace OpenVic2;
 
-TestSingleton *TestSingleton::singleton = nullptr;
+TestSingleton* TestSingleton::singleton = nullptr;
```
Spartan322, line 9: Foremost disagrees with https://github.com/OpenVic2Project/OpenVic2/blob/537938683b748dbf5fcd78276aa823f168f715f1/extension/src/TestSingleton.hpp#L11

Also the type is part of the variable declaration in C++, the pointer is attached to variable identifier so should be associated with the identifier over the type.

Spartan322 (changes requested):
Format changes should be separated from this PR, format changes/fixes should be delegated to either a separate commit or PR, as it also reduces possibility of conflict by modifying unrelated files unnecessarily so. This probably should be part of our contribution guide.

extension/src/TestSingleton.cpp
ClarkeCode: >Format changes should be separated from this PR, format changes/fixes should be delegated to either a separate commit or PR, as it also reduces possibility of conflict by modifying unrelated files unnecessarily so.

>Also the type is part of the variable declaration in C++, the pointer is attached to variable identifier so should be associated with the identifier over the type.

Until the style guide is done, all formatting is considered temporary

>Should also separate declarations into headers and definitions into source files whenever possible.

In general, I would agree with this sentiment. However the purpose of the PR is to give a more advanced example of Godot UI interacting with C++ code, so it is in a single file for ease as an instructional tool

extension/src/Simulation.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <godot_cpp/classes/object.hpp>
+#include <godot_cpp/core/class_db.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+#include <vector>
+
+namespace OpenVic2 {
+	class Simulation : public godot::Object {
+		std::vector<size_t> exampleProvinces;
+
+		//BEGIN BOILERPLATE
```
KaiserOfNone, line 12: I'm not a fan of inheritance, but if we're gonna use a bunch of singletons, we should have a `Singleton` subclass of `godot::Object` that does most of this work. Just leaving `_bind_methods` to be implemented since that one changes a lot. Or a template, that works too.

extension/src/TestSingleton.cpp
Spartan322: > Until the style guide is done, all formatting is considered temporary

Format changes (to the main repo) should still be separate form any PR outside of a format fix PR, this should be enforced as a general guideline default for commits on all version control systems regardless of a styleguide (or a contribution guide either) as this is a very good way to introduce conflict to everyone else's forks without warning. If we have one single commit/PR for format corrections every so often (should only need it once but say this happens again) we can publicly point to it and when its merged have folks update and rebase to it, but if we do it in any PR whenever we can we will have many random format changes constantly.

> In general, I would agree with this sentiment. However the purpose of the PR is to give a more advanced example of Godot UI interacting with C++ code, so it is in a single file for ease as an instructional tool

Well foremost we should adhere to the principals of best formatting rules that make sense where possible while we lack a styleguide, so even beside my previous point regarding formatting commits, a best rule for this case would be to keep the pointer with the identifier, this would teach best practices because even the little details of formatting will become default for how people will write the code even before the styleguide is written. This should be properly addressed in the completed styleguide but default manner of writing should follow how the C++ parser will read it.

Spartan322 (changes requested):

extension/src/Simulation.hpp
ClarkeCode: I agree that this would be the way to go.

I don't imagine that we would need much more than the core simulation to be a singleton since it can provide an interface that the Godot frontend uses

extension/src/Simulation.hpp
```diff
@@ -0,0 +1,52 @@
...
+		GDCLASS(Simulation, godot::Object);
```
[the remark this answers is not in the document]
ClarkeCode: As Kaiser has said here: https://github.com/OpenVic2Project/OpenVic2/pull/8#discussion_r1103841167
We would pull out the singleton functionality into a subclass for production code. This is just an example of how GDExtension works.

game/icon.svg.import
```diff
@@ -16,9 +16,9 @@ dest_files=["res://.godot/imported/icon.svg-218a8f2b3041327d8a5756f3a245f83b.cte
 [params]
 
 compress/mode=0
+compress/high_quality=false
```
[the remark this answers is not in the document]
ClarkeCode: I only included this out of an abundance of caution, since it only seems to be data about the default icon SVG it's impact should be low.

extension/src/Simulation.hpp
```diff
@@ -0,0 +1,52 @@
...
+		static Simulation* _simulation;
+
+		protected:
+		static void _bind_methods() {
+			godot::ClassDB::bind_method(godot::D_METHOD("conductSimulationStep"), &Simulation::conductSimulationStep);
+			godot::ClassDB::bind_method(godot::D_METHOD("queryProvinceSize"), &Simulation::queryProvinceSize);
+		}
+
+		public:
+		inline static Simulation* get_singleton() { return _simulation; }
```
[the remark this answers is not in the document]
ClarkeCode: It's good to know about those macros 👍

extension/src/Simulation.hpp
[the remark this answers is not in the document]
Spartan322: This is about the GDCLASS macro, this has no relevance to that and you can't put that in separate class, it needs to be part of the class its in, also iirc the semicolon here is actually a C++ extension and is not part of the C++ standard.

extension/src/TestSingleton.cpp
ClarkeCode: Changes of commit e7ee0dd have been reverted for now

extension/src/Simulation.hpp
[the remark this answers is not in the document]
ClarkeCode: Ah, I thought you meant the whole block of singleton related code.
Yeah the GDCLASS macro should be at the lowest rung of any inheritance chain.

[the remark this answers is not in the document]
ClarkeCode: Moved GDCLASS macro to top of class declaration

Spartan322 (approved):
I honestly hate this PR for format reasons, but whatever I guess we're not going to care about trying to keep this stuff clean when we don't have a styleguide...even though this is VCS contribution thing.

KaiserOfNone (approved):
Do create a ticket for generalizing singletons tho :)

extension/src/Simulation.hpp
realderekstevens: Very cool.

realderekstevens:
Very cool.

PR #10, opened 2023-02-15
Update and refine docs/styleguide-cpp.md

Spartan322:
Aligns better with Godot's C++ style
Advises preference for Godot types with doc links and rationale
Uses tables to designate necessary information on Godot type preferences
Adds a more specific table of contents
Headers are numbered to make references easier

### DIFF
--- a/docs/styleguide-cpp.md
+++ b/docs/styleguide-cpp.md
@@ -1,320 +1,660 @@
-# OpenVic2 C++ Style Guidelines 
+# OpenVic2 C++ Style Guidelines
 
 ## Table of Contents
- 1. [Why Style?](styleguide-cpp.md#1-why-style)
 
-	1.1. [General Principles](styleguide-cpp.md#11-general-principles)
+- [Why Style?](#why-style)
+- [1. Conventions](#1-conventions)
+  - [1.1. General Principles](#11-general-principles)
+  - [1.2. File Formatting](#12-file-formatting)
+  - [1.3. Naming Conventions](#13-naming-conventions)
+  - [1.4. Do’s and Don’ts](#14-dos-and-donts)
+    - [1.4.1. Keep the opening scope brace on the same
+      line](#141-keep-the-opening-scope-brace-on-the-same-line)
+    - [1.4.2. Keep spaces between arithmetic operators, relational
+      operators, logical operators, variables, and
+      literals](#142-keep-spaces-between-arithmetic-operators-relational-operators-logical-operators-variables-and-literals)
+    - [1.4.3. Prefix and postfix operators should be immediately
+      attached to the
+      variable](#143-prefix-and-postfix-operators-should-be-immediately-attached-to-the-variable)
+    - [1.4.4. Avoid `inline`, use Godot’s `_FORCE_INLINE_` for release
+      inline behavior and `_ALWAYS_INLINE_` for always inline
+      behavior](#144-avoid-inline-use-godots-_force_inline_-for-release-inline-behavior-and-_always_inline_-for-always-inline-behavior)
+    - [1.4.5. Avoid function definitions in header files, prefer
+      definitions in implementation
+      files](#145-avoid-function-definitions-in-header-files-prefer-definitions-in-implementation-files)
+    - [1.4.6. The content of classes should appear in the following
+      order:](#146-the-content-of-classes-should-appear-in-the-following-order)
+    - [1.4.7. Class visibility specifiers should be the same indentation
+      of the
+      class](#147-class-visibility-specifiers-should-be-the-same-indentation-of-the-class)
+    - [1.4.8. Class visibility specifiers should be followed by a
+      declaration on a new line at the next indentation
+      level](#148-class-visibility-specifiers-should-be-followed-by-a-declaration-on-a-new-line-at-the-next-indentation-level)
+    - [1.4.9. Methods which do not mutate an object’s state should be
+      marked as a `const`
+      method](#149-methods-which-do-not-mutate-an-objects-state-should-be-marked-as-a-const-method)
+    - [1.4.10. When argument or parameter lists are too long to easily
+      fit on one line: begin the argument list on the next line, two
+      indentation levels higher than the function’s
+      indentation.](#1410-when-argument-or-parameter-lists-are-too-long-to-easily-fit-on-one-line-begin-the-argument-list-on-the-next-line-two-indentation-levels-higher-than-the-functions-indentation)
+    - [1.4.11. In header files, put `#pragma once` at the top of the
+      file, not C-style header
+      guards](#1411-in-header-files-put-pragma-once-at-the-top-of-the-file-not-c-style-header-guards)
+    - [1.4.12. Use enum classes, not plain
+      enums](#1412-use-enum-classes-not-plain-enums)
+    - [1.4.13. Use type aliases, not
+      \`typedef\`s](#1413-use-type-aliases-not-typedefs)
+    - [1.4.14. Use explicit types, do not use
+      `auto`](#1414-use-explicit-types-do-not-use-auto)
+    - [1.4.15. Do not use C-style casts, instead use C++ casts (ie:
+      `static_cast`, `const_cast`,
+      `reinterpret_cast`)](#1415-do-not-use-c-style-casts-instead-use-c-casts-ie-static_cast-const_cast-reinterpret_cast)
+    - [1.4.16. When declaring a pointer/reference, put a space between
+      the type and the asterisk/ambersand, put the asterisk/ambersand
+      immediately before the variable
+      name.](#1416-when-declaring-a-pointerreference-put-a-space-between-the-type-and-the-asteriskambersand-put-the-asteriskambersand-immediately-before-the-variable-name)
+    - [1.4.17. Do not declare multiple pointer variables on a single
+      line](#1417-do-not-declare-multiple-pointer-variables-on-a-single-line)
+    - [1.4.18. Keep the initialization of a variable and its declaration
+      as close as possible, ideally on the same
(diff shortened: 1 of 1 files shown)


docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+9.  Any submodules or third-party utilities should be in the
+    `OpenVic2/extension/deps` directory
+
+10. Groups of related source files should be kept in a sub-directory of
+    `OpenVic2/extension/src`
+
+11. Any lines that are longer than 120 characters in length should be
+    split across multiple lines
+
+12. Use British English spellings and standards.
```
FakeByte, line 138: Why? You argue for following godot more closely which uses American spelling

docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+
+### 1.3. Naming Conventions
+
+|        |                                   |                                     |                                         |
+|--------|-----------------------------------|-------------------------------------|-----------------------------------------|
+| Number | Item                              | Writing Convention                  | Example                                 |
+| 1      | Class names                       | PascalCase                          | MyCoolExample                           |
+| 2      | Struct names                      | PascalCase                          | MyCoolExample                           |
+| 3      | Variable names                    | Lower snake_case                    | my_cool_example                         |
```
FakeByte, line 147: Changing those conventions is totally subjective preference, I would stick with what @ClarkeCode decided on

docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+    //Something cool
+}
+
+//Incorrect
+for (int x = 0; x < 10; x ++ ) {
+    //Something cool
+}
+```
+
+#### 1.4.4. Avoid `inline`, use Godot’s `_FORCE_INLINE_` for release inline behavior and `_ALWAYS_INLINE_` for always inline behavior
```
FakeByte, line 202: What about non godot c++ code?

docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+
+protected:
+    int example_protected_variable;
+
+private:
+    int example_private_variable;
+};
+```
+
+#### 1.4.8. Class visibility specifiers should be followed by a declaration on a new line at the next indentation level
```
FakeByte, line 297: Also totally subjective change

docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+
+```c++
+//Correct
+int my_number = static_cast<int>(some_float_value);
+
+//Incorrect
+int my_number = (int) some_float_value;
+```
+
+#### 1.4.16. When declaring a pointer/reference, put a space between the type and the asterisk/ambersand, put the asterisk/ambersand immediately before the variable name.
```
FakeByte, line 417: Also subjective change

docs/styleguide-cpp.md
```diff
@@ -1,320 +1,660 @@
...
+#include <godot_cpp/classes/object.hpp>
+#include "openvic2/src/SampleHeader2.hpp"
+#include <stdio>
+#include <godot_cpp/core/class_db.hpp>
+#include <vector>
+#include <string>
+#include "Implementation.hpp"
+```
+
+#### 1.4.23. Prefer Godot’s Variant types over Godot’s template types and the Standard Library
```
FakeByte, line 560: Also what about code that doesnt use godot?

FakeByte (changes requested):

docs/styleguide-cpp.md
Spartan322: Previously agreed upon decision for the styleguide.

docs/styleguide-cpp.md
Spartan322: Reliance on Godot is inevitable if you're gonna use a game engine.

docs/styleguide-cpp.md
Spartan322: I honestly don't much care for it either.

docs/styleguide-cpp.md
Spartan322: It distinguished visibility clearly, else it blends in with the rest of the class.

docs/styleguide-cpp.md
Spartan322: Game engine reliance is inevitable for use of a game engine.

docs/styleguide-cpp.md
Spartan322: Consistent standards beg to differ.

docs/styleguide-cpp.md
schombert: Semantically some things in C++ need to be marked inline even if you really don't care if the compiler actually compiles it as inline, and so you explicitly don't want the inlining to be forced. It is also nicer to do constexpr inline ... with global constants to prevent them from being materialized by taking an address

docs/styleguide-cpp.md
schombert: I guess you could declare aliases for anything in Godot with an American spelling ...

docs/styleguide-cpp.md
schombert: Using some of these types, and hence not using the standard library types that do the same things, is going to make it extremely hard to use anything else in the C++ ecosystem

daveynotfound (commented):
lgtm

docs/styleguide-cpp.md
ClarkeCode: > Reliance on a game engine is inevitable if you're gonna use a game engine.

There is a difference between relying on Godot for rendering/sound/UI and making our core simulation dependant on the internals of Godot's engine

For example, if somebody wanted to write an OpenGL UI/frontend for OpenVic2, it would make no sense for the C++ simulation to be reliant on non-standard Godot types

docs/styleguide-cpp.md
ClarkeCode: > Also totally subjective change

@FakeByte is correct, this is merely a matter of preference

docs/styleguide-cpp.md
ClarkeCode: Aside from Godot-specific types/macros polluting the C++ codebase more than necessary, the ability to conditionally inline functions is not necessary.

The rationale given for conditional inlining is that it makes code easier to debug, however inlined methods should only be used for trivial (or near-trivial) attribute accessors.

Reminder that most of the team's developers have intermediate-to-expert levels of C++ experience.

docs/styleguide-cpp.md
ClarkeCode: > Consistency standards beg to differ.

To what consistency standards are you referring?
If you are talking about keeping C++ code consistent with GDscript, there is no extrinsic or functional reason to do so.

This is subjective preference.

docs/styleguide-cpp.md
ClarkeCode: > Previously agreed upon decision for the styleguide.

I do not believe that there was ever such a decision for the *code styleguide*

I don't think it's useful to explicitly mandate the spelling of words
It's literally Poh-tay-to vs Poh-tah-to  ¯\_(ツ)_/¯

docs/styleguide-cpp.md
Spartan322: It says prefer for a reason.

docs/styleguide-cpp.md
Spartan322: This was had as a discussion and vote last month and it was decided that we will rely on British spelling as the default for everything, it was not referring to only localization.

docs/styleguide-cpp.md
Spartan322: And C++ code with other C++ code, it is useless to interface with Godot and violate what it does as it only creates confusion and requires nearly doubly the documentation work and at least twice as much work dealing with error messages.

docs/styleguide-cpp.md
Spartan322: None of this addresses this issue and by now referring to "near-trivial" usage you've given even better justification for why conditional inline matters, if use of inline matters, there is no excuse not to rely on conditional inline as even in trivial cases it will make debug more difficult.

docs/styleguide-cpp.md
Spartan322: Calling it subjective doesn't resolve anything if there is a reason behind the change.

docs/styleguide-cpp.md
FakeByte: > This was had as a discussion and vote last month and it was decided that we will rely on British spelling as the default for everything, it was not referring to only localization, code style was explicitly mentioned. What is the point of a styleguide if it does not explicitly define things consistently? If we're gonna keep saying this is "subjective" then why do we need a styleguide for anything? Code style is inherently subjective, that's a meaningless excuse.

The vote was "for localization purposes no other reason honest, thumb this up if you are British or spell colour like that <——"

docs/styleguide-cpp.md
Hop311: I personally find lower snake case easier to read.

docs/styleguide-cpp.md
Hop311: I also prefer having access specifiers an indent level lower than the definitions following them, it helps break up the code into sections rather than one big block.

docs/styleguide-cpp.md
Hop311: It may be subjective, but my preference is still for `void *ptr`.

docs/styleguide-cpp.md
schombert: The C++ core guidelines (https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines) is directing people to underscore_style these days, and strongly discourages ALL_CAPS for enum constants. Just FYI

ClarkeCode:
Certain formatting things may be integrated at a later date, current version is sufficient for now

PR #15, opened 2023-02-21
Add Keychain plugin for Controls tab

Spartan322:
Adds Controls tab

See https://github.com/Orama-Interactive/Keychain/tree/4.x

Keychain README:
> Keychain is a plugin for the Godot Engine that aims to give the player full control over the input actions of the game. 
>
> ### Features
>
> - Easy to use shortcut re-mapping system, for all input types, keyboard, mouse, gamepad buttons and gamepad axes. Great for accessibility purposes.
> - Add multiple input events (shortcuts) to each action, of multiple input types.
> - Group your inputs together, and choose whether certain input actions are global or local. Global events cannot be re-assigned anywhere else, local events can, but not to other events in the same group.
> - Choose between different shortcut profiles and create custom ones.
> - Save shortcut modifications.
> - Support for localization.
> - Automatically deletes old input actions that have been deprecated during development of your project, so there is no need to worry over this.

Keychain is covered under the MIT license and I have included the license in the addons folder.

In Godot, plugins are convenient engine/game functionality developed for the sake of more generalized use to speed up some element of development. This is just as much a demonstration of plugins as it is a useful fulfillment of controls functionality through Godot. It also demonstrates common standards and good practices in Godot

Since this is directly embedded into the project as well there is no issue modifying the plugin to suit our needs even better or to correct bugs.

Obviously it should still be tested to ensure correct functionality according to requirements.

Note that if you have previously run the project and had backed out of the settings in your openvic2 user directory, this is incompatible with that old settings.cfg and thus it must be deleted.

This also implements a Godot engine project override polish to account for lack of godotengine/godot-proposals#6247, but only in binary export else it screws with the project.godot file.

https://user-images.githubusercontent.com/4315446/220351156-c784cd08-3625-4a79-8465-bd64af850d0f.mp4

https://user-images.githubusercontent.com/4315446/220351159-0f419e9c-5f88-4c30-b267-48f86439fc1d.mp4


ClarkeCode (approved):
Looks like a good plugin to use for rebindable keystrokes

PR #16, opened 2023-02-21
Add Credits System

Orwellian-225:
Added a data based Credits system that will search for a credits.csv file found within res://common/ and then render the credits according to the csv file.

The csv file format is:
title,project-title
role-name,person-name
role-name,person-name
...

Roles will be capitalized in application so can be lowercase.

The only needed row in the file for it to be accepted is title,project-title. All other rows are dynamically read and added.

Any additional data included in a row will not be read so it will not throw an error 

Requirements met by this PR:
* SS-17
* SS-29 [1]
* UI-32
* UI-34
* UIFUN-36
* UI-35
* UI-38
* UIFUN-37
* FS-4

Footnotes:
[1]: This requirement is sort of satisfied. It is technically possible for the system to render additional credit files, it does not have any functionality to find them. This is because the mod structure has not been fully mapped out yet, so its not possible to write this code without it almost certainly being redundant immediately. It will however accept and render multiple files once they have been found

### DIFF
--- a/game/src/Credits/Credits.gd
+++ b/game/src/Credits/Credits.gd
@@ -1,0 +1,116 @@
+extends Control
+
+signal back_button_pressed
+
+###############
+# Credits CSV format
+# The project title row is the only requirement within the csv file, however
+# it can be on any row, so long as it exists.
+# ----------------------
+# title,project-title
+# role-name,person-name
+# role-name,person-name
+# role-name,person-name
+# ...
+###############
+
+@export_file("*.csv")
+var core_credits_path : String
+
+# TODO: implement for theme instead
+# waiting for https://github.com/OpenVic2Project/OpenVic2/pull/48
+@export_group("Label Settings", "label_settings_")
+@export
+var label_settings_project : LabelSettings
+
+@export
+var label_settings_role : LabelSettings
+
+@export
+var label_settings_personel : LabelSettings
+
+@export
+var credits_list: VBoxContainer
+
+# REQUIREMENTS:
+# * 1.5 Credits Menu
+# * SS-17
+
+# REQUIREMENTS
+# * FS-4
+func load_credit_file(path):
+	var roles = {}
+	var core_credits = FileAccess.open(path, FileAccess.READ)
+
+	while not core_credits.eof_reached():
+		var line = core_credits.get_csv_line()
+
+		# If the line does not have an identifiable role or is empty then skip it
+		if line[0].is_empty():
+			continue
+
+		var role := line[0].to_upper()
+		var person : String = line[1] # Cannot infer with := as line[1] does not have a set type
+
+		if role not in roles.keys():
+			roles[role] = []
+			roles[role].push_back(person)
+		else:
+			roles[role].push_back(person)
+
+	return roles
+
+# REQUIREMENTS:
+# * UI-34, UI-35
+func make_project_credits(project):
+	var project_credits_list = VBoxContainer.new()
+	var project_credits_label = Label.new()
+
+
+	# Spartan has some suggestions here but for now its good enough. Refer
+	# to PR 16 resolved comments for further details
+	project_credits_list.name = 'Credits' + project['TITLE'][0]
+	project_credits_label.name = 'Label' + project['TITLE'][0]
+	project_credits_label.text = project['TITLE'][0]
+	project_credits_label.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+	project_credits_label.label_settings = label_settings_project
+
+	for role in project.keys():
+		if role == 'TITLE':
+			continue
+
+		var role_parent = VBoxContainer.new()
+		var role_label = Label.new()
+
+		role_parent.name = 'Role' + role
+		role_label.name = 'Label' + role
+		role_label.text = role
+		role_label.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+		role_label.label_settings = label_settings_role
+
+		for person in project[role]:
+			var person_label = Label.new()
+
+			person_label.name = 'Label' + person
+			person_label.text = person
+			person_label.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+			person_label.label_settings = label_settings_personel
+			role_parent.add_child(person_label)
+
+		project_credits_list.add_child(role_label)
+		project_credits_list.add_child(role_parent)
+		project_credits_list.add_child(HSeparator.new())
+
+	credits_list.add_child(project_credits_list)
+
+# REQUIREMENTS:
+# * SS-17
+func _ready():
+	var core_credits = load_credit_file(core_credits_path)
+	make_project_credits(core_credits)
+
+# REQUIREMENTS:
+# * UI-38
+# * UIFUN-37
+func _on_back_button_pressed():
+	back_button_pressed.emit()
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ bin/*
 
 # MacOS stuff
 .DS_Store
+
+*.translation
--- a/game/common/credits.csv
+++ b/game/common/credits.csv
@@ -1,0 +1,43 @@
+Title,OpenVic2
+Project Lead,FakeByte
+Project Lead,joethepro36
+Project Lead,Catylist
+Project Lead,patrick c
+Project Lead,ZincLadder
+Vice Lead,hop311
+Community Manager,Catylist
+Project Management,JackRalph96
+Developer,Dekaizer
+Developer,forn
+Developer,ItsLateHere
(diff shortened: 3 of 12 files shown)


game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+# it can be on any row, so long as it exists.
+# ----------------------
+# title,project-title
+# role-name,person-name
+# role-name,person-name
+# role-name,person-name
+# ...
+###############
+
+@export var coreCreditsPath : String = 'res://common/credits.csv'
```
Spartan322, line 17:
```suggestion
@export_file("*.csv")
var core_credits_path : String = 'res://common/credits.csv'
```

This should probably be a custom CSVLoader resource instead to be honest, but I'll leave that for later implementation.

game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+			continue
+		
+		var role: String = line[0].to_upper()
+		var person: String = line[1]
+		
+		if role not in roles.keys():
+			roles[role] = []
+			roles[role].push_back(person)
+		else:
+			roles[role].push_back(person)
```
Spartan322, line 43:
```suggestion
	while not coreCredits.eof_reached():
		var line := coreCredits.get_csv_line()
		
		# If the line does not have an identifiable role or is empty then skip it
		if line[0].is_empty():
			continue
		
		var role := line[0].to_upper()
		var person := line[1]
		
		if role not in roles.keys():
			roles[role] = []
			roles[role].push_back(person)
		else:
			roles[role].push_back(person)
```
line, role, and person should be able to be inferred

game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+# * UI-34, UI-35
+func drawProjectCredits(project):
+	var projectParent = VBoxContainer.new()
+	var projectLabel = Label.new()
+	
+	projectParent.name = 'Credits-' + project['TITLE'][0]
+	projectLabel.name = 'Label-' + project['TITLE'][0]
+	projectLabel.text = project['TITLE'][0]
+	projectLabel.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+	projectLabel.label_settings = load("res://src/Credits/ProjectLabelSettings.tres")
```
Spartan322, line 57: Construction should not be performed here in GDScript? 

This only requires `projectLabel.text = project['TITLE'][0]` but even for that I would encourage separating the title from the csv, for now leaving it in the scene anyway. The rest of this should be part of the Credits scene. In the future if we're gonna put the modders in this list, a csv would only list the role and name, the project name will be handled separately.

game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+		roleLabel.label_settings = load("res://src/Credits/RoleLabelSettings.tres")
+		
+		for person in project[role]:
+			var personLabel = Label.new()
+			
+			personLabel.name = 'Label-' + person
+			personLabel.text = person
+			personLabel.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+			personLabel.label_settings = load("res://src/Credits/PersonLabelSettings.tres")
+			roleParent.add_child(personLabel)
```
Spartan322, line 79: More efficient to make this its own VBoxContainer scene so its actually possible to manage this from the GUI editor, avoid construction to the scene tree unless it cannot be managed using scene instantiation, if you have questions about that, ask me and I'll demonstrate it.

game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+		
+		projectParent.add_child(roleLabel)
+		projectParent.add_child(roleParent)
+		projectParent.add_child(HSeparator.new())
+	
+	$"Scroll/Projects".add_child(projectLabel)
+	$"Scroll/Projects".add_child(projectParent)
```
Spartan322, line 86: Avoid relying on hardcoded scene paths, instead you should have a
```gdscript
@export
var projects_node : Control
```
that you assign in the scene editor so that you can refactor the scene in the scene editor without issue and one can reassign the scene if it gets deleted or lost somehow. Avoid hardcoded paths and strings.

game/src/Credits/Credits.gd
```diff
@@ -0,0 +1,98 @@
...
+			continue
+		
+		var roleParent = VBoxContainer.new()
+		var roleLabel = Label.new()
+		
+		roleParent.name = 'Role-' + role
+		roleLabel.name = 'Label-' + role
+		roleLabel.text = role
+		roleLabel.horizontal_alignment = HORIZONTAL_ALIGNMENT_CENTER
+		roleLabel.label_settings = load("res://src/Credits/RoleLabelSettings.tres")
```
Spartan322, line 70: Disregarding the next review comment, you should prefer preload over load as it takes no runtime cost, but also you shouldn't even be hardcoding the string if you can help it, you should always prefer export variables like so:

```gdscript
@export
var role_label_settings : LabelSettings
# ...
roleLabel.label_settings = role_label_settings
```

game/project.godot
```diff
@@ -23,7 +23,7 @@ Resolution="*res://src/Autoload/Resolution.gd"
 
 window/size/viewport_width=1280
 window/size/viewport_height=720
-window/size/mode=3
+window/size/initial_position_type=2
```
Spartan322, line 26:
```suggestion
window/size/mode=3
```

Spartan322 (changes requested):
The .translation file should be removed from this PR, they need to be added to the .gitignore as they are a kind of local cache file for CSV files.

I need to open this PR up in Godot, but I believe the credits.csv.import file isn't supposed to be a translation type. I'll have to check on that.

The gd files should follow the Godot standard style with lower snake_case member variables, signals, and functions. (This PR is already style inconsistent because its relying on that in the same files that's it violating it in)

game/project.godot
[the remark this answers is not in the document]
Orwellian-225: Because I play with settings. I'll revert it don't worry. Just figuring out what godot does and doesn't use

game/src/Credits/Credits.gd
Orwellian-225: They can be inferred, I was just being explicit. Also I know is_empty is correct it was just giving me issues(I need to get better at comments sorry)

game/src/Credits/Credits.gd
Orwellian-225: Moders credits was agreed to be a separate CSV file found in the mod directory, hence the need for a title within the CSV to identify it

Orwellian-225: However theoretically that title could be thrown into a some sort of meta data file and handled separately but it wasn't important for now

game/src/Credits/Credits.gd
Orwellian-225: Yeah sure. We probably need to hop on a call or something so I can see

game/src/Credits/Credits.gd
Orwellian-225: Will do, it was already giving me issues when I was fiddling with the display so I'll clean that up as well

game/src/Credits/Credits.gd
Orwellian-225: Will take a look at it, thanks

game/project.godot
Orwellian-225: Yeah sorry, I play with settings

Orwellian-225:
> The .translation file should be removed from this PR, they need to be added to the .gitignore as they are a kind of local cache file for CSV files.
> 
> 
> 
> I need to open this PR up in Godot, but I believe the credits.csv.import file isn't supposed to be a translation type. I'll have to check on that.
> 
> 
> 
> The gd files should follow the Godot standard style with lower snake_case member variables, signals, and functions. (This PR is already style inconsistent because its relying on that in the same files that's it violating it in)
> 
> 
> 
> 

Yeah I'll clean it up. I wasn't sure what was happening so I used the gd standard where there was other code and the cpp style in the stuff I wrote from scratch. I just wasn't sure what was happening and defaulted to the cpp

game/src/Credits/Credits.gd
Orwellian-225: This is good enough for now I think. We'll need to come revisit this anyway once the artists have something in mind so we can take a look at it then

Spartan322:
This has conflicts, it should probably be rebased, the conflicts would still need to be resolved.

Spartan322:
Alright squash and rebase complete and conflicts are resolved.

Spartan322 (approved):

ClarkeCode (approved):

Hop311:
Fixes #35

PR #17, opened 2023-02-22
Delete Simulation singleton correctly.

Hop311:


Spartan322 (approved):
Yeah better to avoid leaking memory whenever possible.

PR #21, opened 2023-02-25
Add Setting Sanatization

Hop311:
- Settings input sanatization for Locale, Resolution, Screen Mode, Monitor, Audio sliders.
- Revamped Resolution handling system to better handle custom resolutions.
- Various debug and release checks to avoid unexplained crashes.

### DIFF
--- a/game/src/LocaleButton.gd
+++ b/game/src/LocaleButton.gd
@@ -1,5 +1,8 @@
 extends OptionButton
 
+const section_name : String = "Localization"
+const setting_name : String = "Locale"
+
 var _locales_country_rename : Dictionary
 var _locales_list : Array[String]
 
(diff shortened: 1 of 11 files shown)


game/src/LocaleButton.gd
```diff
@@ -21,15 +24,32 @@ func _ready():
...
+	if file == null: return
+	file.set_value(section_name, setting_name, _locales_list[selected])
+
+func reset_setting() -> void:
+	selected = _locales_list.find(TranslationServer.get_locale())
+
+func _on_item_selected(index : int) -> void:
+	if _valid_index(index):
+		TranslationServer.set_locale(_locales_list[index])
+		Events.Options.save_settings_to_file.call_deferred()
```
Spartan322, line 52: One thing that is strange with this is that it will perform a double save if the player is in the options menu, the effects of this won't cause any issues right now, it'll only ever crop up to the player if we were to remove saving on application close and application crash (or in the case Godot is not running with a crash handler) but it is funky behavior in my opinion.

Spartan322 (approved):
Looks good, I don't see anything problematic sticking out.

Spartan322:
> joethepro36: please add requirements references if you can as comments for functions please

Since this is modifying the settings functionality anyway, where applicable this would probably be a good addition for these specific modified files anyway.

PR #23, opened 2023-02-25
Basic SFX Engine

BrickPi:
Built out a very basic sound effect engine. I have a few thoughts about how exactly "play" should be called but didn't include in this PR because it would've mandated bigger changes to scene root.


Spartan322 (changes requested):
The keychain assets import files should be removed from the commit.

Spartan322:
Since audio playback is a bit special and quirky with Godot, I might examine a more robust manner of dealing with it soon.

Spartan322:
I wasn't saying this had to be closed necessarily.

PR #25, opened 2023-02-25
Add Sound Effect Manager

BrickPi:
Supercedes #23 

Basic engine for sound effects, unimpacted by loading new scenes.


game/src/MainMenu/MainMenu.gd
```diff
@@ -12,23 +12,28 @@ func _ready():
 
 
 func _on_new_game_button_pressed():
+	await get_node("/root/GameMenu/SFX").play("click")
 	print("Start a new game!")
 	get_tree().change_scene_to_file("res://src/SampleGame.tscn")
 
 
 func _on_continue_button_pressed():
+	get_node("/root/GameMenu/SFX").play("click")
```
schombert: This is going to be a problem eventually. You want a more uniform, and less error prone, way of making the click sound play for all button presses without having to manually add a command to each button action.

Spartan322 (changes requested):
The hardcoded nature of this is quite a problem.

The lack of autoload for what is a pretty good case for an autoload singleton scene with a script is also an issue.

Do not call parent or sibling scenes and nodes from child classes, parent classes call down the node tree, a node should never call upwards or sideways. We should always adhere to the principal of call down (as in the parent can call the children methods) and signal up. (the children call their signals which are connected up to the parent) This is the principal of the observer pattern Godot uses with signals as it prevents intercoupled scene structures that can't be refactored.

game/src/MainMenu/MainMenu.gd
CptAlanSmith: No it's not

BrickPi: I completely agree, but that's not my purpose with this PR; I'm adding a way to play sound effects, not a way to automagically call it when buttons are pressed.

schombert: Since this is a common thing (add sounds to button clicks) there is probably some way built into Godot to handle it without manually adding the sound in the function body

BrickPi:
@Spartan322
Refactored to use autoload and should be good now regarding calling parents/siblings.

schombert:
Suggest looking at: https://www.reddit.com/r/godot/comments/hrit58/add_a_sound_to_all_the_buttons_in_a_project/

game/src/Autoload/SFX.gd
```diff
@@ -0,0 +1,20 @@
+extends Node
+
+var loadedSFX = {}
+
+func _ready():
+	var dir = DirAccess.open("res://audio/sfx/")
+	for fname in dir.get_files():
+		if fname.get_extension() == "ogg":
+			loadedSFX[fname.split(".")[0]] = load("res://audio/sfx/" + fname) # SND-10
```
Spartan322, line 9: split will fail produce weird results if a file happens to have a period in it, prefer `get_basename()`
```suggestion
			loadedSFX[fname.get_basename()] = load("res://audio/sfx/" + fname) # SND-10
```

game/src/Autoload/SFX.gd
```diff
@@ -0,0 +1,20 @@
+extends Node
+
+var loadedSFX = {}
```
Spartan322, line 3: Should probably be `_loaded_sounds`

Spartan322 (changes requested):
Otherwise should be good, I might investigate something more sophisticated (and usable) later.

game/src/Autoload/SFX.gd
BrickPi: done in 60d8b55

game/src/Autoload/SFX.gd
BrickPi: done in 60d8b55

game/src/GameMenu.tscn
```diff
@@ -38,5 +38,4 @@ alignment = 2
 [node name="LocaleButton" parent="HBox" instance=ExtResource("4_jno35")]
 layout_mode = 2
 
-[connection signal="options_button_pressed" from="MainMenu" to="." method="_on_main_menu_options_button_pressed"]
```
Spartan322, line 41: This would prevent the OptionsMenu from ever showing if you clicked the options button. Pretty sure this should be reverted.

Spartan322 (changes requested):

game/src/GameMenu.tscn
BrickPi: should be fixed in dba52cd, I wonder why it did that

Spartan322 (approved):

Spartan322 (approved):

PR #26, opened 2023-02-26
Add skeleton for Game Lobby Menu

nuruvilu:
The lobby is opened by clicking "new game" in the main menu. Currently only two start dates are in the listbox, but the intention is for saved games to be present as well. Most of the space on the right-hand panel is reserved for information about the selected country and other players. The purpose of the separation is to allow a map to be rendered behind the menu.
The start button can be pressed only after a save has been selected.

Save games/start dates don't actually get loaded yet, since obviously the actual underlying game data doesn't exist yet.

image

### DIFF
--- a/game/src/LobbyMenu/LobbyMenu.gd
+++ b/game/src/LobbyMenu/LobbyMenu.gd
@@ -1,0 +1,27 @@
+extends Control
+
+signal back_button_pressed
+signal save_game_selected
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	pass # Replace with function body.
+
+
+func _on_back_button_button_down():
+	print("Returning to Main Menu.")
+	back_button_pressed.emit()
+
+
+func _on_start_button_button_down():
+	print("Starting new game.")
+	get_tree().change_scene_to_file("res://src/SampleGame.tscn")
+
+
+func _on_game_select_list_item_selected(index):
+	print("Selected save game: ", index)
+	save_game_selected.emit(index)
+
+
+func _on_save_game_selected(_index):
+	$GameStartPanel/VBoxContainer/StartButton.disabled = false
--- a/game/src/LobbyMenu/LobbyMenu.gd
+++ b/game/src/LobbyMenu/LobbyMenu.gd
@@ -1,27 +1,35 @@
 extends Control
 
+# REQUIREMENTS:
+# * 1.4 Game Lobby Menu
+# * SS-12
+
 signal back_button_pressed
 signal save_game_selected
 
-# Called when the node enters the scene tree for the first time.
-func _ready():
-	pass # Replace with function body.
+@export
+var start_button : BaseButton
 
-
+# REQUIREMENTS:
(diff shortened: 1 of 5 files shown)


game/src/LobbyMenu/LobbyMenu.gd
```diff
@@ -0,0 +1,27 @@
...
+	get_tree().change_scene_to_file("res://src/SampleGame.tscn")
+
+
+func _on_game_select_list_item_selected(index):
+	print("Selected save game: ", index)
+	save_game_selected.emit(index)
+
+
+func _on_save_game_selected(_index):
+	$GameStartPanel/VBoxContainer/StartButton.disabled = false
```
Spartan322, line 27: Prefer an export variable to hardcoded node paths. Example:
```gdscript
@export
var start_button : BaseButton
# ...
start_button.disabled = false
```
This makes scene refactoring trivial.

game/src/LobbyMenu/LobbyMenu.gd
```diff
@@ -0,0 +1,27 @@
+extends Control
+
+signal back_button_pressed
+signal save_game_selected
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	pass # Replace with function body.
+
```
Spartan322, line 9: Since its not used:
```suggestion
```

game/src/LobbyMenu/LobbyMenu.gd
```diff
@@ -0,0 +1,27 @@
...
+
+func _on_back_button_button_down():
+	print("Returning to Main Menu.")
+	back_button_pressed.emit()
+
+
+func _on_start_button_button_down():
+	print("Starting new game.")
+	get_tree().change_scene_to_file("res://src/SampleGame.tscn")
```
Spartan322, line 18: Prefer an export variable to a hardcoded scene file path. Example:
```gdscript
@export
var game_scene : PackedScene
# ...
get_tree().change_scene_to_packed(game_scene)
```

Spartan322 (changes requested):
Seems functionally good, few things need to be changed, also would prefer to have dev cycle requirements as comments, if you're unsure I'll see about dealing with it myself in a later PR anyway.

game/src/LobbyMenu/LobbyMenu.gd
nuruvilu: For some reason using an export variable here caused the game to get stuck on the load screen, so I've left it out for now.

nuruvilu:
Added requirements tags.

game/src/LobbyMenu/LobbyMenu.gd
Spartan322: I believe this is a case of godotengine/godot#73983, a regression introduced in Godot 4 RC 4, should be fixed in Godot 4 RC 6, I'll test it beforehand to make sure.

Spartan322 (approved):
Looks good.

PR #43, opened 2023-02-26
Update Keychain SVG imports to Godot 4 RC 5

Spartan322:
A change happened in the release candidates for Godot that changed the compression settings for texture imports


Spartan322:
> Did they change it again with RC6?

This is a permanent change, it appears this change was introduced a month ago with godotengine/godot#72031 and I just ignored it on the keychain assets because keychain itself wasn't update to date with the release candidates. If you examine the icon.svg.import, you'll see this is how its supposed to be.

BrickPi (approved):
Right sounds good.

ClarkeCode (approved):

PR #44, opened 2023-02-27
Main Menu Version and Checksum Labels

BrickPi:
What it says on the tin, as well as the boilerplate for a C++ hook that's called to actually calculate the checksum. Currently reports checksum "1234abcd". Also added proper tabstop/focus for the main menu buttons.

Satisfies UI-31 & UI-111


Spartan322 (approved):
We should probably prefer separating our headers into declaration headers and source definition files. This is fine for now, otherwise looks fine.

PR #45, opened 2023-02-27
Refactor SFX into a SoundManager

Spartan322:
Add SoundManager able to play arbitrary sound streams Make SoundManager use only one AudioStreamPlayer per bus
Add StyleBoxWithSound
Add Kenney UI Audio click3.wav
Removed sound play via pressed signals in MainMenu
Make Button_MainMenu pressed style StyleBoxWithSound with click3.wav sound


game/src/Autoload/SoundManager.gd
```diff
@@ -0,0 +1,36 @@
...
+
+var _loaded_sound : Dictionary = {}
+
+var _bus_to_stream_player : Dictionary = {}
+
+func _ready():
+	var dir = DirAccess.open(_audio_directory_path)
+	for fname in dir.get_files():
+		match fname.get_extension():
+			"ogg", "wav":
```
BrickPi, line 13: If we're going to support ogg and wav we might as well load mp3's too if only for the flexibility.

game/src/Utility/StyleBoxWithSound.gd
```diff
@@ -0,0 +1,33 @@
...
+	get: return sound
+	set(v):
+		sound = v
+		emit_changed()
+
+func _get_draw_rect(rect : Rect2) -> Rect2:
+	if style_box == null: return Rect2()
+	return style_box._get_draw_rect(rect)
+
+func _draw(to_canvas_item : RID, rect : Rect2) -> void:
```
BrickPi, line 23: Interesting solution certainly, not entirely sure about meshing audio with draw functionality but it works for now.

BrickPi (changes requested):
Looking good, would add mp3 as accepted file type.

game/src/Utility/StyleBoxWithSound.gd
Spartan322: Yeah, its kinda the only way to get audio functionality into the styles, which is easier to manage then trying to inject into every pressed signal to output a sound somehow.

BrickPi (approved):
Looks good, I'm gonna approve but it doesn't look like github acknowledges it.

ClarkeCode (approved):
Seems good!

PR #48, opened 2023-02-27
Make default_theme a text resource

Spartan322:
Change default_theme.theme to default_theme.tres: Makes it a text resource
Remove Button font data from default_theme: Fixes erroneous theme updates
Adds Label_Checksum.font_color theme color, replaces ChecksumLabel's theme override value.


BrickPi (approved):
Looking good!

PR #52, opened 2023-03-02
Make LobbyMenu resolution independent and fix MacOS export directory

Spartan322:


BrickPi (approved):
Good work, looks good as far as I can see.

PR #56, opened 2023-03-05
Added the ability to select a save file

H12324:
Added the ability to select a save file if it is added to the SaveSytem folder and has a .ov2 extension and created a save button which creates a save called savegame.ov2. Though the save files themselves are currently meaningless and there is no load functionality.

### DIFF
--- a/game/src/LobbyMenu/GameSelectList.gd
+++ b/game/src/LobbyMenu/GameSelectList.gd
@@ -1,0 +1,34 @@
+extends ItemList
+
+# REQUIREMENTS:
+# * SS-18
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	var files = get_saved_files()
+	print(files)
+	for file in files:
+		self.add_item(file.get_basename())
+
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(delta):
+	pass
+
+func get_saved_files():
+	var dir = DirAccess.open("res://src/SaveSystem") # Will store saves here for now refactor later
+	var files = []
+	
+	if dir:
+		dir.list_dir_begin()
+		var file_name = dir.get_next()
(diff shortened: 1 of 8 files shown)


game/src/SaveSystem/SaveSystem.gd
```diff
@@ -0,0 +1,18 @@
+extends Node
+
+# Called when the node enters the scene tree for the first time.
+func _ready():
+	pass # Replace with function body.
+
+
+# Called every frame. 'delta' is the elapsed time since the previous frame.
+func _process(delta):
+	pass
```
Spartan322, line 10:
```suggestion
extends Node
```

game/src/LobbyMenu/GameSelectList.gd
```diff
@@ -0,0 +1,34 @@
...
+		dir.list_dir_begin()
+		var file_name = dir.get_next()
+		while file_name != "":
+			if !dir.current_is_dir():
+				print("Found file: " + file_name)
+				if file_name.get_extension() == "ov2":
+					files.append(file_name)
+			file_name = dir.get_next()
+	else:
+		print("An error occurred when trying to access the path.")
```
Spartan322, line 31: Should probably prefer `push_error`

Spartan322 (changes requested):
We should avoid using print more.

Should prefer the `user://` directory for save management (in like a `user://save_games` or something)

This doesn't create the save directory if it doesn't exist.

`.ov2` is implied to be a binary file isn't it?

Probably also shouldn't save things within the game's source files.

If you need a global access singleton, you should be using singleton autoloads with Godot.

Saves should also be able to be named.

Generated comments and functions should be removed.

Prefer export variables is possible to hardcoded values, avoid hardcoded values that affect behavior as much as possible.

PR #57, opened 2023-03-09
Cleanup MusicPlayer for future development

Spartan322:
Add music import files
Rename MusicUIController to MusicPlayer
Adjust MusicPlayer node position, offset, and anchors
Adjust MusicConductor, MusicPlayer, and SongInfo style to conform with other GDScript files
Correct capability for MusicPlayer to desync from the MusicConductor
Adjusted MusicPlayer button text to use media player unicode symbols
Adjust MusicPlayer to correct mouse filter problems
Adjust MusicPlayer to appear more consistent
Correct lack of path_join use in SongInfo
Correct mouse filter problems in OptionsMenu


KaiserOfNone (approved):
FTR you can use `git mv` to move or rename a file and not delete all the git history (and well spam the diffs)

PR #58, opened 2023-03-09
Added Data modelling Documentation - JC and Vanilla

Orwellian-225:
add: ideology documentation
add: goods documentation
add: sim doc format

I've added some data modelling to the documentation. This only covers partially covers the ideology and goods for now, but its a base that allows us to start actually keeping track of these things.

Couple notes:
This doesn't reference the Requirements in anyway and there are some pieces of data in here that I don't think are in reqs. This should probably be amended sometime in the future

### DIFF
--- a/docs/simulation/goods.md
+++ b/docs/simulation/goods.md
@@ -1,0 +1,146 @@
+# OpenVic2 Goods
+
+## Victoria 2 Structure
+
+## OpenVic2 Structure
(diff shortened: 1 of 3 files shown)


ClarkeCode (dismissed):
Looks great! Well done!

ClarkeCode (approved):

PR #59, opened 2023-03-10
Add Splash Startup functionality

Spartan322:
Add boot splash screen
Implement splash skip on interaction

Note the video demonstrates the functionality, not the implementation.

https://user-images.githubusercontent.com/4315446/224229472-b422bcb0-1497-4a94-8213-3bd9487a901d.mp4

Bit of a janky implementation to handle the video, but it does at least work and looks fine.

`VideoStreamPlayer`s can take a few frames to initialize video streams meaning its invisible for at least two frames, and they don't have capability to pause on video end (you can't even recognize how close the video is to completion either) so the `VideoStreamPlayer` becomes invisible. To fix this two images are needed, a simulation of the first frame and the last frame, thus hiding the janky implementation of `VideoStreamPlayer` from the user.

If not in requirements, perhaps should be added to requirements?


game/src/GameStart.tscn
```diff
@@ -0,0 +1,52 @@
...
+[ext_resource type="VideoStream" path="res://splash_assets/splash_startup.ogv" id="5_8euyy"]
+
+[node name="GameStartup" type="Control"]
+layout_mode = 3
+anchors_preset = 15
+anchor_right = 1.0
+anchor_bottom = 1.0
+grow_horizontal = 2
+grow_vertical = 2
+
```
BrickPi, line 16: I feel like this is a good place to replace the whole tree, rather than having the startup screen still loaded in memory.

BrickPi (changes requested):
Other than that nitpick it looks pretty good, if you don't want to change it I'm planning on implementing a full loading screen and I'll roll it into that refactor.

game/src/GameStart.tscn
Spartan322: The implementation I made is the cheapest manner to a implement startup splash without running the splash screen when loading the game menu scene back in memory. Otherwise you'd need to account for another scene.

BrickPi (approved):

PR #68, opened 2023-03-22
Docs architecture and schema

ClarkeCode:


orwellian225 (dismissed):
All seems good at the moment. I'd like to note though we've discussed the the changing of both the Dataloader as well as how the Sim passes data back and forth between the Client. I'd still recommend merging this in so the original plan is recorded in the docs at some point

ClarkeCode:
> All seems good at the moment. I'd like to note though we've discussed the the changing of both the Dataloader as well as how the Sim passes data back and forth between the Client. I'd still recommend merging this in so the original plan is recorded in the docs at some point

These docs will definitely be subject to revisions and refinement over time 👍

Hop311 (approved):
LGTM, great work on the mermaid diagrams :)

PR #69, opened 2023-03-22
Added type header

ClarkeCode:


FakeByte:
Why would you remove the size of the type from its name? I think we will end up using 32 and 64 bit types and having the name clearly specify their size will simplify it.

Spartan322:
Why are you using `std::string`? This isn't the simulation stuff and Godot can't serialize `std::string`.

ClarkeCode:
Taking this back to the drawing board

PR #74, opened 2023-03-24
Fix Export Issues

Spartan322:
Fix MacOS OpenVic2.dmg export_path to OpenVic2.app
Rename macOS export preset to MacOS
Add `common/*` to include_filter
Make credits.csv not translation file but to keep the file on import


ClarkeCode (approved):
Seems good to me 👍

PR #75, opened 2023-03-24
Add extensive continuous integration and artifacts 

Spartan322:
Relies on #74 

Add push artifacts and release asset uploading
Add release integration
Add pck debug check and artifact
Update CI workflow with release archives
Update CI workflow for all pushes
Remove initial compile of godot-cpp from workflow

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -1,6 +1,6 @@
 name: Builds
 
-on: [pull_request]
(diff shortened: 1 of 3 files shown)


Spartan322:
This is now confirmed to work, it speeds up the build process and produces a binary for windows, mac, and linux that can be immediately opened to confirm functionality.

Issue #31, opened 2023-02-26
SS-61 not documented


Spartan322:
If a language is not available it shall fall back to the en_GB locale

Spartan322:
Fixed by #77

Issue #30, opened 2023-02-26
SS-60 not documented


Spartan322:
If a translation for a certain locale isn't implemented it shall fall back to the language's default locale

Spartan322:
Fixed by #77

PR #77, opened 2023-03-25
Added Localisation

Hop311:
- Replaced hard coded button/label text with keys localised in localisation CSVs
- Added localisation loading system as described in reqs
- Added default locale selection as described in reqs
- Removed unnecessary imports (credits.csv was incorrectly being imported as a translation)
- standardised settings sections and names to snake case keys rather than localised/display text

Fulfils: SS-57, SS-59, SS-60, SS-61, FS-17, FS-18, FS-23, FS-24, FS-25


ClarkeCode (dismissed):
Looks beautiful!

Spartan322 (dismissed):
I really don't agree with how you're doing localization here, I'll explain in a few.

Spartan322:
First off this is not just about localization, that gdignore in common will just break exports. We should be exporting all our data, if we want to extract them, we can do that on first run, but we should not be requiring an external folder to be placed in the user's directory, this creates a massive amount of stupid problems.

Spartan322:
Second off, this makes the process of localizing with Godot a massive hassle, LoadLocalization is completely unnecessary here, all of that can be done with already in Godot, at minimum the Translation resource already allows this, you can already externally load resources anyway.

Spartan322:
Also what even is the point of using a CSV file for translations if you're gonna do this?

Hop311:
I wrote a custom translation CSV loader in order to fit the specifications in the requirements, mainly that all translations in a folder called `xx_YY` should automatically apply to that locale, as the default CSV translation importer, even when the CSV files were formatted correctly, would sometimes remove the locale (e.g. have the name "English, " instead of "English, <locale>") and generally didn't seem to handle entries for the same locale spread across multiple files. There were also more minor reasons like the default CSV translation importer crashing for certain inputs, rather than gracefully exiting with an error code, e.g. for an empty file.

With the `.gdiginore`s I didn't realise they prevented exporting, I'll remove them and manually disable auto-importing for things which should be exported but not auto-imported (e.g. `credits.csv`).

As for `gettext` and `.po` files, they look interesting, nobody had brought them up at the last couple meetings discussing dev cycle 2 tasks (which this localisation implementation was one of), but thanks for mentioning them now I'll take a look.

Spartan322:
See #78 for a more dynamic localization system, I suggest using that as the basis and adapting the translation tags from here to that, also `.po` files have special editors you can use for them.

Hop311:
I'll first try to fix the issues with this PR while still implementing what is specified in the requirements. If you think we need to change the localisation requirements with respect to file format and structure, please bring it up in a meeting or in a text channel so everyone can discuss the proposed redesign.

Spartan322:
> I'll first try to fix the issues with this PR while still implementing what is specified in the requirements. If you think we need to change the localisation requirements with respect to file format and structure, please bring it up in a meeting or in a text channel so everyone can discuss the proposed redesign.

I couldn't be at any of the meetings when this was planned, there was no way for me to contribute to that, and I don't think its strictly necessary for me to point it out in the chat when an implementation can be examined to accomplish the same thing better, this aside it appears like nobody actually examined Godot's localization capability when building the requirements, and the requirements specification on use CSVs were quite naive. Also the common standard term is localization with a z, most modern software uses a z, not an s, even Paradox does this, Victoria 2 uses the s, but at least since Crusader Kings 3, they use localization. Regardless I think at minimum my changes should be examined as the basis, if necessary I can modify it to allow separation of localization files by locale, however with how po files, it wouldn't actually accomplish anything that prevents someone from putting say a British English file in a France French locale directory. (though I could have the system complain if you do that I suppose)

ClarkeCode (dismissed):

Spartan322:
In the least could this be moved out of the extensions and into GDScript, there's no reason it should be kept in C++.

Spartan322:
Fixes #31

Spartan322:
Fixes #30

PR #78, opened 2023-03-26
Add runtime and resource import localization

Spartan322:
Supercedes #77

Removed `_locales_list = [TranslationServer.get_locale()]` from LocaleButton.gd

This does not implement the translations themselves nor the translation id strings like with #77, I suggest a separate PR for those, however this makes it trivial to add translations and use Godot generated translations. (note Godot forgets to escape `"` in its pot generator, if someone wants to report that as a bug at https://github.com/godotengine/godot before I do, be my guest) This also supports runtime translations overriding normal game localization. As presented this does not implement a display in the editor that appears translated, but I suppose we'd consider that a secondary necessity anyway. I recommend that this be tested on Windows and Mac to confirm that it will properly handle runtime translations on other platforms too, its unlikely to now, I just wish to be sure.

One note I should make is that TranslationLoader should be the left as the last autoload, it should need to be accessed, so its not a global variable autoload.

### DIFF
--- a/game/project.godot
+++ b/game/project.godot
@@ -26,6 +26,7 @@ MusicConductor="*res://src/MusicConductor/MusicConductor.tscn"
 SoundManager="*res://src/Autoload/SoundManager.gd"
 Keychain="*res://addons/keychain/Keychain.gd"
 GameDebug="*res://src/Autoload/GameDebug.gd"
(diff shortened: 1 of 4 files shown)


ClarkeCode:
See dev discussion

PR #87, opened 2023-04-13
Add style configs

andrewmustea:
I would have opened a discussion thread but there doesn't seem to be one available on this project.

The formatting for cpp files is all over the place and we should use clang-format to standardize it.

Config taken from here: https://github.com/godotengine/godot-cpp/blob/master/.clang-format

We could also use a well known standard like LLVM.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,202 @@
+# Parameters are equivalent to the Godot cpp style.
+# See https://github.com/godotengine/godot-cpp/blob/master/.clang-format
+#
+# Commented out parameters are those with the same value as base LLVM style.
+# We can uncomment them if we want to change their value, or enforce the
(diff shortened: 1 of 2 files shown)


Spartan322:
Honestly this is the style I would prefer to use for this project.

Spartan322:
As an aside @andrewmustea do you think you could add a simple formatting check action to this PR?

andrewmustea:
> As an aside @andrewmustea do you think you could add a simple formatting check action to this PR?

Yes I can do that.

ClarkeCode:
Thank you @andrewmustea for bringing `clang-format` to our attention, it seems like it will be quite useful.
I've made a PR to add a `.clang-format` file to the project: #95

If you'd like to chat more about the project, you'll find us here: https://discord.gg/3xcH8xZB

PR #90, opened 2023-04-16
C++ Style Tweaks

ClarkeCode:
Tweaked C++ files to conform to C++ styleguide

### DIFF
--- a/extension/src/Checksum.hpp
+++ b/extension/src/Checksum.hpp
@@ -11,7 +11,7 @@ namespace OpenVic2 {
 
 	protected:
 		static void _bind_methods() {
-			godot::ClassDB::bind_method(godot::D_METHOD("get_checksum_text"), &Checksum::get_checksum_text);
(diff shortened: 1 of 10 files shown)


Hop311:
I'm kinda torn on how to handle naming, I had prioritised properly descriptive names which wasn't an issue with snake case, but that doesn't work well camel case, e.g.
```
get_province_index_from_uv_coords vs get_province_identifier_from_uv_coords
```
becomes
```
getProvinceIndexFromUVCoords vs getProvinceIdentifierFromUVCoords
```
which IMO are harder to differentiate at a glance. 
Also if we change the C++ version we should do the same for the Godot version, using different names for the same function introduces an extra place something can go wrong for no real gain.
```
ClassDB::bind_method(D_METHOD("get_province_index_from_uv_coords", "coords"), &MapSingleton::getProvinceIndexFromUVCoords);
ClassDB::bind_method(D_METHOD("get_province_identifier_from_uv_coords", "coords"), &MapSingleton::getProvinceIdentifierFromUVCoords);
```

ClarkeCode:
Will be revisited at a later time

PR #95, opened 2023-04-23
Added a .clang-format for formatting use

ClarkeCode:
This `.clang-format` will be used in the future for automatic formatting.
The auto-formatting will be addressed between dev cycle 2 and 3.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,48 @@
+---
+UseCRLF: false
+Standard: c++20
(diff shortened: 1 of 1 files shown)


Hop311 (approved):
Looking good 😎

Spartan322:
@ClarkeCode have you even tested this? Cause this format is still bugged, it won't actually ever run?

PR #97, opened 2023-04-24
Support features up to clang-format 14

Spartan322:
Uses IndentExternBlock
Uses AccessModifierOffset
Use EmptyLineBeforeAccessModifier
Use IndentAccessModifiers
Use PackConstructorInitializers
Use IncludeBlocks
Use IncludeCategories

Update C++ files within `extension/src` to follow .clang-format


.editorconfig
```diff
@@ -0,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = tab
+#insert_final_newline = true
```
Hop311, line 7: I'm happy to have this enabled.

Hop311 (commented):
`.clang-format` and `.editorconfig` look good, the PR just needs the formatted C++ files to be updated, and after merging we can do the same for the submodule.

Hop311 (approved):
Looks good, the `src/openvic2` stuff will have to be re-formatted in the submodule once `create-simulation-submodule` is merged but that shouldn't be an issue.

ClarkeCode (approved):

PR #101, opened 2023-04-28
Dynamic minimap + vertical map subdivision

Hop311:
This moves the terrain byte (currently 0 for water and 1 for land) from the province-colour-lookup texture to the province-index texture array, to allow terrain type to differ within a province without needing a separate texture array.
Horizontal subdivision counts are calculated as the smallest positive divisor `n` of `width`, such that `width / n < 16384`, with vertical subdivisions calculated similarly. This could have problems with certain numbers, e.g. primes over `16384` will result in the dimension being divided into individual pixels.


Spartan322:
I really don't get why you're messing with GameDebug here, I've got a few ideas on refining it, that into an argument parser which stores argument values in a `openvic2/arguments` ProjectSettings category.

Hop311:
> I really don't get why you're messing with GameDebug here, I've got a few ideas on refining it, probably making it into an argument parser which stores argument values in a `openvic2/arguments` ProjectSettings category.

My understanding was that `Events` is for global singletons which don't need to be a `Node` and can just be `RefCounted` instead, like `Options`. Is there a different reason for `Options` not being a `Node`?

ClarkeCode (approved):
Implementation works, gj 👍

Spartan322:
@Hop311 Events is a global eventing singleton, and anything related to global events, signals and management for global signals basically. It has nothing to do with that, the reason I set it up the way was for namespace separation purposes.

PR #103, opened 2023-04-30
Cosmetic terrain + Hashmaps

Hop311:
- Added a system for loading additional terrain textures and identifying which texture to use where based on an image, similar to how province shapes are defined.
- Introduced hashmaps to replace linear searches.
- Made the currently selected province get deselected when the Province Overview Panel is closed.
- Made the current hover province reset when the mouse is over a menu.
- Highlight selected province on minimap + fixed clicking on minimap.


ClarkeCode (approved):
Great work, love the minimap highlighting!

PR #104, opened 2023-05-01
Better province highlight on selection/hover.

Hop311:
- Province highlight on selection/hover is now a 10% shift to white for each, rather than an increase in the proportion of the province colour used vs the terrain colour.
- Province Overview Panel now blocks mouse hover over provinces, just like Map Control Panel
- Made `pixels_per_terrain_tile` a constant, eventually this should take both the map dims and the terrain texture dims into account.


Spartan322 (approved):
Looks good.

ClarkeCode (approved):

PR #105, opened 2023-05-02
Fix display names for North American provinces

ClarkeCode:
Tweak localization keys so that names will show up properly on the province overview panel

### DIFF
--- a/game/localisation/en_GB/provinces.csv
+++ b/game/localisation/en_GB/provinces.csv
@@ -48,48 +48,48 @@ prov_ukraine,Ukraine
 prov_russia,Russia
 
 ,, North America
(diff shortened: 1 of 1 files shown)


Hop311:
Could you fix the rest of the North America names? Not sure why they still have `_NAME` when everything else doesn’t, probably some commits getting merged in a weird order

ClarkeCode:
> Could you fix the rest of the North America names? Not sure why they still have `_NAME` when everything else doesn’t, probably some commits getting merged in a weird order

Sure thing 👍

Hop311 (approved):
Thanks for cleaning this up 🙂

PR #106, opened 2023-05-03
Moved simulation code to separate submodule

Hop311:
- Simulation module is in `extension/deps/openvic2-simulation`
- SConstruct takes a list of paths instead of just `extension/src`
- Updated `README` to mention starting twice to import assets correctly.


ClarkeCode (approved):
Great work with the submodule

PR #111, opened 2023-05-10
Added Goods loading + mapmode

Hop311:
- This also moves province selection to the C++ side so it can be used in mapmodes, simplifying it with a single `province_selected` signal belonging to `GameSingleton`.
- Defines loading is reorganised into a single function that takes a `Dictionary` of filepaths, rather than having to call each load function from GDScript individually.
- Fixed buildings in `ProvinceOverviewPanel` getting destroyed and recreated on update.
- Fixed wrong import for terrain textures + improved `load_terrain_variant_file` error handling.
- Switched to better string formatting for `LIFE_RATING_TOOLTIP` and `MAINMENU_CHECKSUM`.

Requirements completed:
- MAP-23
- UI-552, UI-561
- UIFUN-131


Spartan322 (approved):
Seems good running on release build artifact, nothing sticks out as problematic, and builds completely fine for all platforms, so looks good to merge.

PR #113, opened 2023-05-15
Add Keep Resolution Dialog

Nemrav:
Fix GameSession UI node order

- Fulfills requirements UI-873, UIFUN-301, UIFUN-302
- Adds a dialog popup to the videotab under the options menu. This dialog appears when the user selects a new resolution, asking them if they want to keep the resolution. Selecting 'No' or letting the dialog's timer expire will revert the resolution, while selecting 'yes' will keep the newly selecting resolution.
- Localized, but french isn't my native language so it needs to be double checked.
- Motivation: if the user messes up their resolution, they can now recover gracefully.
- Also: Changed GameSessionMenu to be lower in GameSession's hierarchy so that it appears on-top of any overlapping UI elements (achievable with a bad UI-scaling setting).
- The KeepChangesDialog.tscn is designed to be somewhat reusable if needed.

### DIFF
--- a/game/localisation/en_GB/menus.csv
+++ b/game/localisation/en_GB/menus.csv
@@ -36,6 +36,8 @@ OPTIONS_VIDEO_REFRESH_RATE,Refresh Rate
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Only change from VSYNC if you are having issues with screen tearing.
 OPTIONS_VIDEO_QUALITY,Quality Preset
 OPTIONS_VIDEO_GUI_SCALE,Gui Scaling Factor
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TITLE,Keep Resolution?
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TEXT,Reverting resolution in ({time})
 
 ,, Sound Tab
 MASTER_BUS,Master Volume
--- a/game/localisation/en_US/menus.csv
+++ b/game/localisation/en_US/menus.csv
@@ -36,6 +36,8 @@ OPTIONS_VIDEO_REFRESH_RATE,Refresh Rate
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Only change from VSYNC if you are having issues with screen tearing.
 OPTIONS_VIDEO_QUALITY,Quality Preset
 OPTIONS_VIDEO_GUI_SCALE,Gui Scaling Factor
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TITLE,Keep Resolution?
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TEXT,Reverting resolution in ({time})
 
 ,, Sound Tab
 MASTER_BUS,Master Volume
--- a/game/localisation/fr_FR/menus.csv
+++ b/game/localisation/fr_FR/menus.csv
@@ -36,6 +36,8 @@ OPTIONS_VIDEO_REFRESH_RATE,Taux de Rafraîchissement
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Ne changez de VSYNC que si vous rencontrez des problèmes de déchirement d'écran.
 OPTIONS_VIDEO_QUALITY,Préréglage de la Qualité
 OPTIONS_VIDEO_GUI_SCALE,Mise à échelle de l'Interface Graphique
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TITLE,Maintenir la résolution?
+OPTIONS_VIDEO_RESOLUTION_DIALOG_TEXT,Reverserai la résolution en ({time})
 
 ,, Sound Tab
 MASTER_BUS,Volume Principal
--- a/game/src/OptionMenu/ResolutionSelector.gd
+++ b/game/src/OptionMenu/ResolutionSelector.gd
@@ -3,9 +3,15 @@ extends SettingOptionButton
 # REQUIREMENTS
 # * UIFUN-21
 # * UIFUN-28
+# * UIFUN-301
+# * UIFUN-302
 
 @export
 var default_value : Vector2i = Resolution.error_resolution
+
+var previous_resolution : Vector2i = Resolution.error_resolution
+@export var revert_dialog : ConfirmationDialog
+@export var timer : Timer
 
 func _find_resolution_index_by_value(value : Vector2i) -> int:
 	for item_index in item_count:
@@ -63,7 +69,31 @@ func _set_value_from_file(load_value):
 
 func _on_item_selected(index : int):
 	if _valid_index(index):
+		previous_resolution = Resolution.get_current_resolution()
 		Resolution.set_resolution(get_item_metadata(index))
+		var new_resolution = get_item_metadata(index)
+		
+		#has_focus() indicates the user is calling _on_item_selected, not some other function
+		if has_focus() and previous_resolution != new_resolution:
+			print("Start Revert Countdown!")
+			start_revert_countdown()
 	else:
 		push_error("Invalid ResolutionSelector index: %d" % index)
 		reset_setting()
+
+func _process(_delta):
+	revert_dialog.dialog_text = tr("OPTIONS_VIDEO_RESOLUTION_DIALOG_TEXT").format({"time":round(timer.time_left)})
+
+func start_revert_countdown() -> void:
+	timer.start()
+	revert_dialog.popup_centered(Vector2(1,1))
+	
+func _on_confirmed() -> void:
+	timer.stop()
+
+func _cancel_changes() -> void:
+	Resolution.set_resolution(previous_resolution)
+	_sync_resolutions()
+	print("Resolution reset to (%dx%d)" % [previous_resolution.x,previous_resolution.y])
+	timer.stop()
+	revert_dialog.hide()
--- a/game/src/OptionMenu/VideoTab.tscn
+++ b/game/src/OptionMenu/VideoTab.tscn
@@ -31,7 +31,7 @@ columns = 2
 layout_mode = 2
 text = "OPTIONS_VIDEO_RESOLUTION"
 
-[node name="ResolutionSelector" type="OptionButton" parent="VBoxContainer/GridContainer"]
+[node name="ResolutionSelector" type="OptionButton" parent="VBoxContainer/GridContainer" node_paths=PackedStringArray("revert_dialog", "timer")]
 editor_description = "UI-19"
 layout_mode = 2
 focus_neighbor_bottom = NodePath("../ScreenModeSelector")
@@ -40,8 +40,22 @@ selected = 0
 popup/item_0/text = "MISSING"
 popup/item_0/id = 0
 script = ExtResource("1_i8nro")
+revert_dialog = NodePath("ConfirmationDialog")
+timer = NodePath("Timer")
 section_name = "video"
 setting_name = "resolution"
+
+[node name="ConfirmationDialog" type="ConfirmationDialog" parent="VBoxContainer/GridContainer/ResolutionSelector"]
+editor_description = "UI-873"
+disable_3d = true
+title = "OPTIONS_VIDEO_RESOLUTION_DIALOG_TITLE"
+size = Vector2i(730, 100)
+ok_button_text = "DIALOG_OK"
+cancel_button_text = "DIALOG_CANCEL"
+
+[node name="Timer" type="Timer" parent="VBoxContainer/GridContainer/ResolutionSelector"]
+wait_time = 5.0
+one_shot = true
 
 [node name="GuiScaleLabel" type="Label" parent="VBoxContainer/GridContainer"]
 layout_mode = 2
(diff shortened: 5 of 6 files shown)


game/src/OptionMenu/KeepChangesDialog.gd
```diff
@@ -0,0 +1,53 @@
...
+	#if revert_callback is somehow null, you should see a console error
+	revert_callback.call(old_setting)
+	timer.stop()
+	hide()
+	
+func _process(_delta):
+	var time_left = -1
+	if not timer.is_stopped():
+		time_left = timer.time_left
+	dialog_text = tr("REVERTING_IN") % time_left
```
Spartan322, line 37: Please do not rely on % for formatting, see [this discord message](https://discord.com/channels/1063392556160909312/1064542494114725909/1105107520441307237). Prefer using [`format`](https://docs.godotengine.org/en/4.0/tutorials/scripting/gdscript/gdscript_format_string.html#format-method-examples) and try to prefer format naming so purpose is more clear.

game/localisation/en_GB/menus.csv
```diff
@@ -36,6 +36,10 @@ OPTIONS_VIDEO_REFRESH_RATE,Refresh Rate
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Only change from VSYNC if you are having issues with screen tearing.
 OPTIONS_VIDEO_QUALITY,Quality Preset
 OPTIONS_VIDEO_GUI_SCALE,Gui Scaling Factor
+REVERTING_IN,Reverting resolution in (%d)
```
Spartan322, line 39: Avoid %d.

game/localisation/en_US/menus.csv
```diff
@@ -36,6 +36,10 @@ OPTIONS_VIDEO_REFRESH_RATE,Refresh Rate
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Only change from VSYNC if you are having issues with screen tearing.
 OPTIONS_VIDEO_QUALITY,Quality Preset
 OPTIONS_VIDEO_GUI_SCALE,Gui Scaling Factor
+REVERTING_IN,Reverting resolution in (%d)
```
Spartan322, line 39: Avoid %d.

game/localisation/fr_FR/menus.csv
```diff
@@ -36,6 +36,10 @@ OPTIONS_VIDEO_REFRESH_RATE,Taux de Rafraîchissement
 OPTIONS_VIDEO_REFRESH_RATE_TOOLTIP,Ne changez de VSYNC que si vous rencontrez des problèmes de déchirement d'écran.
 OPTIONS_VIDEO_QUALITY,Préréglage de la Qualité
 OPTIONS_VIDEO_GUI_SCALE,Mise à échelle de l'Interface Graphique
+REVERTING_IN,Reverserai la résolution en (%d)
```
Spartan322, line 39: Avoid %d.

game/src/OptionMenu/VideoTab.gd
```diff
@@ -2,8 +2,27 @@ extends HBoxContainer
 
 @export var initial_focus: Control
 
+@onready var revert_dialog = get_node("KeepChangesDialog")
+@onready var resolution_selector = get_node("VBoxContainer/GridContainer/ResolutionSelector")
```
Spartan322, line 6: Please do not rely on `get_node`, always prefer to use export nodes when you need reference things, as its less prone to breaking on refactoring and it does not require messing with code to fix it via Godot's interface.

Spartan322 (changes requested):
I would recommend that instead of having a KeepChangesDialog scene and script, that you'd stick all that into a ResolutionSelector script and put everything in there from the KeepChangesDialog for now, since the only thing you can revert is resolution. Instead what you should do is put the dialog and timer under the ResolutionSelector node in the VideoTab and reference them in the script via export variables only when necessary. For composition reasons, always avoiding putting functionality of nodes at the top level of scenes, as this enables a decoupled workflow enables us to streamline refactoring and development, no need to think about the the interconnected nature of the project.

Reason to avoid the reusability thing is that we should avoid expanding the scope of implementation details like this unless know we'll use it somewhere else, and in that case it would be trivial to refactor it for this purpose, keep things otherwise specific however to their purposes at as narrow a scope as possible so that we can compose things independently.

Spartan322:
Also try to keep titles concise to most important changes, additional fixes can be mentioned in the issue summary, a reordering of nodes correction is rather insignificant to the functionality and should be more of a footnote, not the issue title.

game/src/OptionMenu/VideoTab.tscn
```diff
@@ -154,7 +155,14 @@ section_name = "video"
 setting_name = "quality_preset"
 default_selected = 1
 
+[node name="KeepChangesDialog" parent="." instance=ExtResource("8_cpkub")]
+title = "KEEP_CHANGES"
+ok_button_text = "YES"
+cancel_button_text = "NO"
```
Spartan322, line 161: Does not follow the de factor standard of the other ConfirmationDialog.

game/src/OptionMenu/VideoTab.tscn
```diff
@@ -40,9 +40,21 @@ selected = 0
...
 script = ExtResource("1_i8nro")
+revert_dialog = NodePath("ConfirmationDialog")
+timer = NodePath("Timer")
 section_name = "video"
 setting_name = "resolution"
 
+[node name="ConfirmationDialog" type="ConfirmationDialog" parent="VBoxContainer/GridContainer/ResolutionSelector"]
+editor_description = "UI-873"
+disable_3d = true
+title = "OPTIONS_VIDEO_RESOLUTION"
```
Spartan322, line 51: Prefer separate titles for dialogs in menus.csv, if you prefer the title not overflow, please keep it really short so it likely does not overflow please. Also I would recommend you use `OPTIONS_VIDEO_RESOLUTION_DIALOG_TITLE` for that.

game/src/OptionMenu/VideoTab.tscn
```diff
@@ -40,9 +40,21 @@ selected = 0
...
+size = Vector2i(730, 100)
+ok_button_text = "OPTIONS_VIDEO_RESOLUTION_DIALOG_OK"
+cancel_button_text = "OPTIONS_VIDEO_RESOLUTION_DIALOG_CANCEL"
```
Spartan322, line 54: Its probably fine to use DIALOG_OK and DIALOG_CANCEL here

Spartan322 (changes requested):
Looks close to good, just a few details to fix up.

game/src/OptionMenu/ResolutionSelector.gd
```diff
@@ -63,7 +71,47 @@ func _set_value_from_file(load_value):
...
+		time_left = timer.time_left
+	revert_dialog.dialog_text = tr("OPTIONS_VIDEO_RESOLUTION_DIALOG_TEXT").format({"time":time_left})
+
+func start_revert_countdown() -> void:
+	timer.start(countdown_seconds)
+	revert_dialog.popup_centered(Vector2(1,1))
+
+func _on_confirmed() -> void:
+	timer.stop()
+	revert_dialog.hide()
```
Spartan322, line 110: This is unnecessary when a dialog's `dialog_hide_on_ok` toggle is on. (which is default for most dialogs)

game/src/OptionMenu/ResolutionSelector.gd
```diff
@@ -63,7 +71,47 @@ func _set_value_from_file(load_value):
...
+
+func _cancel_changes() -> void:
+	Resolution.set_resolution(previous_resolution)
+	_sync_resolutions()
+	print("Resolution reset to (%dx%d)" % [previous_resolution.x,previous_resolution.y])
+	timer.stop()
+	revert_dialog.hide()
```
Spartan322, line 117: This is unnecessary when a dialog's dialog_hide_on_ok toggle is on. (which is default for most dialogs)

game/src/OptionMenu/ResolutionSelector.gd
```diff
@@ -63,7 +71,47 @@ func _set_value_from_file(load_value):
...
+	
+func _on_selection_changed(previous_resolution, _new_resolution):
+	if revert_dialog:
+		print("Start Revert Countdown!")
+		start_revert_countdown()
+		
+func _process(_delta):
+	var time_left:int = -1
+	if not timer.is_stopped():
+		time_left = timer.time_left
```
Spartan322, line 101: This is unnecessary, `time_left` returns 0 if inactive anyway.

game/src/OptionMenu/ResolutionSelector.gd
```diff
@@ -63,7 +71,47 @@ func _set_value_from_file(load_value):
...
 		push_error("Invalid ResolutionSelector index: %d" % index)
 		reset_setting()
+
+#Dialog revert functionality below here
+#UIFUN-301 and UIFUN-302
+
+#This is just to filter out cases where _on_item_selected is not called by the
+#user changing a setting (which happens on startup)
+var was_pressed:bool = false
+func _on_toggled(button_pressed):
```
Spartan322, line 90: `button_pressed` is already a variable in BaseButton, please avoid shadowing variables as it creates warnings in the editor. (and can create problems down the line, hence why its a warning)

Spartan322 (changes requested):
The dialog does not seem to appear when the resolution is set to 720.

I don't think you should be using a node timer for this, I think it be better to use a [SceneTreeTimer](https://docs.godotengine.org/en/stable/classes/class_scenetreetimer.html) instead.

I would also not recommend using signals in the the script that's emitting the signal, instead it be better to have a separate function which sets everything and then emitting a signal there if you're gonna do that. Not exactly sure what the point of the `selection_changed` signal would be here otherwise, but its not a big deal either way.

I would also recommend if possible to either put requirement documentation according to the methods it relates to or at the top of the script using this standard:
```gdscript
# Requirements
# * <req>
# * <req 2>
```

Also you should prefer to put the export Nodes below export variables for neatness, as well you should try to be more specific then Node for export nodes.

game/src/OptionMenu/ResolutionSelector.gd
Nemrav: From experimentation, this seems necessary since this function is triggered both when the timer stops (which means the dialog needs to be hidden manually), and the dialog's cancel option (which requires the timer be stopped.

Nemrav:
> Maybe you shouldn't be using node timers, might be better to use a [SceneTreeTimer](https://docs.godotengine.org/en/stable/classes/class_scenetreetimer.html) instead. But SceneTreeTimers are one-shot, so when they end, they stop processing. If do use node timers, first off set it to one-shot then you don't need to stop it when it reaches zero, secondly you don't need to set a time to the timer if its `wait_time` is already 5 seconds.

I temporarily did a partial reimplementation to use SceneTreeTimer, but I found the Timer node made things somewhat cleaner in the code. I can still switch to it if you strongly prefer.

game/src/OptionMenu/ResolutionSelector.gd
Spartan322: Cancel should be hiding the dialog as well.

Spartan322: One way this could work would be if you could emit the `pressed` signal on `ConfirmationDialog.get_cancel_button()` for the Timer's timeout, you can't do this in the scene but you could set this in the `_ready` function. If there are more places where this pops up, I think I'll come back to write a `TimedConfirmationDialog` node to handles this instead, but for now it shouldn't matter.

Spartan322 (approved):
I guess this is fine enough for now, kinda would prefer to use `var identifier: TypeHint = value` or `var identifier : TypeHint = value` (notice the spaces) but otherwise this should work, its not that big a deal because our style is all over the place anyway. That `has_focus` case might be a bit iffy, but I guess its fine as of now. I might modify this solution a bit later in the case a more generic solution is needed (specifically for timed dialog) but I don't think there is any justification for implementing that now. (and honestly I'd prefer to implement it as a GDExtension element honestly if I did)

Spartan322 (changes requested):
If you would or could, I'd ask for you to squash your commits actually, if you don't know how to, I could guide you on it by the way.

Spartan322:
> With github, I believe that's just a dropdown option once I'm approved to do the merge correct?

This destroys the commit and branch history as it doesn't produce a PR merge commit, what I mean is for you to do a squash so that all the commits, which only exist to fix up the first commit to your PR branch, it is most beneficial to have a track record of PR merge commits and individual PR commits which track the branches in the commit history.

> The reason for having it is ultimately that _on_item_selected() gets called during the splash screen (I'm guessing its load_setting() > _set_value_from_file() > _syncResolution() > _on_item_selected()). If there were a way to have the splash screen start at the saved resolution, or debounce that startup change, it would be unnecessary.

What do you mean by splash screen? Cause the start up animation only really plays after the loading so that's not true. But the startup image can't be controlled by the game at all, the only way to set the resolution for it is by messing with the project settings so the next time it loads it'll start at the correct resolution. (there is godotengine/godot-proposals#6247 which would address this, but there is nothing in terms of implementation for Godot right now, so its not even likely to see in 4.1) Which is something I have already covered in #15, however I found this not useful for debug purposes because it just will override the the project settings in `godot.project` which means it'll set the values in the project. So it only does this on export, and it can't help but save something for that before the game is closed else it'll just use the default project settings.

Nemrav:
> What do you mean by splash screen?

Sorry, I meant the startup image. From the description, it sounds like changing the project settings wouldn't solve this, since for most user settings, the startup image would still be different from the animation onwards. I guess lets leave this.

> If you would or could, I'd ask for you to squash your commits actually, if you don't know how to, I could guide you on it by the way.

I've never done a squash before, from looking it up, is this correct?:
1. Count the number of commits in the branch (in this case 9)
2. `git rebase -i HEAD~9`
3. In the window that pops up, set all commits to "squash" instead of "pick"
4. `git push`

ClarkeCode:
> Count the number of commits in the branch (in this case 9)
> git rebase -i HEAD~9
> In the window that pops up, set all commits to "squash" instead of "pick"
> git push

You are correct, that will squash the branch. If git complains that "local and remote" diverge, you can force push with `git push -f` to overwrite this remote branch

Spartan322:
> Sorry, I meant the startup image. From the description, it sounds like changing the project settings wouldn't solve this, since for most user settings, the startup image resolution would still be different from the animation onwards. I guess lets leave this.

That's why the splash image is the first frame of the animation, there is literally nothing you can do to fix this in Godot because splash images are ran just after runtime initialization and before pretty much any user code could ever be run, its so far ahead it runs before platform initialization code, which means its not even technically specifically aware of what platform the engine is running on when the splash screen is set (unless its a mobile platform which has to do something differently) The only way to control the splash image is through ProjectSettings, nothing else.

Nemrav:
I did the rebase and renamed the commit.

Spartan322 (changes requested):
You seemed to have reverted back to older commits, the issues discussed before need to once again be fixed.

Nemrav:
shoot, I picked the old instead of the new commit, I'll look into fixing this

Spartan322 (approved):
The commit message is in a very strange format, and you left mention of fixes in the commit message despite the fact there is no distinction between the initial commit and commits that fix the initial commit, but these are minor so its fine for a merge.

In future commits it would be a bit better if you would make them more descriptive and readable on what they add to the main branch, that way its less distracting and more transparent to read the commit history.

PR #115, opened 2023-05-16
Fix SettingHSlider not loading properly on default values

Spartan322:
Fix MusicConductor ignoring MUSIC_BUS

This fixes volume sliders being ignored when loaded at zero, which means if you set the slider to zero and reloaded the game, it would act as if the volume was at 100.


Hop311 (approved):
LGTM 👍

Spartan322:
will merge once the checks complete

PR #119, opened 2023-05-19
Pie chart

Nemrav:
Adds a pie/donut chart control node
Adds a temporary chart test scene, which I intend to delete later
- Path to associated nodes is res://src/Utility/PieChart/
- Piechart is made using a shader on a textureRect.
- This is set as `@tool` so a preview is available in a limited capacity in the editor
Controls:
- In the inspector, the "donut" and "donut inner radius" values can only be controlled through the PieChart properties, not the shader.
- The size/outer radius is controlled through the Control > Layout > Transform > size property or the Control > Layout > custom Minimum Size
- The trim size, and whether a donut has an inner radius trim, as well as the gradient base and falloff for the center spotlight are controlled through the CanvasItem > Material > Piechart_shader > Shader Parameters
- Values of the chart can only be added or removed through functions. However, in the editor, you can edit the shader parameters "Stop Angles" and "Colours" to do a mockup.

Areas of concern:
- Some properties are set using `_ready()` such as the shader, as I'm unsure of how to force instances of the PieChart node to use the PieChart shader every time a new PieChart is instanced. I need to change this in the future so that path changes don't break the node.
- Material is duplicated on every instance by code: this is because the shader "instance uniform" feature isn't available in 2d yet. This wastes resources.
-  I'm unsure how to get mouse events to pass through an outer donut chart to an inner pie chart (control elements don't seem to have any way to conditionally pass an input event). 
- What folder should the pie chart stuff go into? for now I'm using utility/PieChart, but this likely needs to change.
- Is there a way I'm missing to lock the size of the node to be square? 

TODO:
- Fix the above if possible
- Validate inputs if needed, else delete the commented out validation function
- Remove Test scene once the chart is ready
- Remove the default pie chart slices, so that pie charts don't need to have them programmatically every time, once testing is done.


Spartan322:
Might want to checkout how https://github.com/fenix-hub/godot-engine.easy-charts/blob/main/addons/easy_charts/control_charts/plotters/pie_plotter.gd does it.

Nemrav:
> Might want to checkout how https://github.com/fenix-hub/godot-engine.easy-charts/blob/main/addons/easy_charts/control_charts/plotters/pie_plotter.gd does it. (be aware that's doing it in Godot 3.x, principals should be the same and the general idea should still apply, but there are definitely references that have changed in Godot 4.0)

The commits before "shader piechart" used this method (godot's _draw() call to make a 2d polygon representation of the chart), so anyone who wants to take a look can check out the last of those commits.
Reason I decided against it was that, to make the chart look nice (gradient falloff), you need to use a shader either way, and godot's shaders don't work well with _draw() from what I read, so we might as well use the shaders the whole way.

An alternative approach I didn't take was another way of doing the gradient: I instead of a function to calculate falloff, I could have blended the piechart with a texture.

Nemrav:
Just a note for myself: Before final merge, the default labels should be removed.

Nemrav:
Broken with Godot 4.1, new pr for pie charts in godot 4.1 is located at https://github.com/OpenVicProject/OpenVic/pull/146

PR #125, opened 2023-05-25
Import fix + Compatibility mode command line option

Hop311:
- Added godot-cpp helper function `load_godot_image` which at runtime can loads real filesystem files and imported files in the `pck` (method used determined by `res://` prefix).
- Changed lots of files from `keep` import to `Image` import.
- Added `--compatibility-mode` command line option to specify vic2 install directory.
- Small changes to `ArgumentParser.gd`:
    - made it a singleton so it can be loaded before `Events.gd`
    - made it so `--key val` options work, before it would be interpreted as `--key="" val`
    - some error handling so invalid inputs won't cause a crash


Spartan322 (approved):
Looks good, that one change in `SongInfo.gd` fixed all the memory leak instances too, including the JSON ones.

PR #131, opened 2023-06-10
Localisation and UI focus cleanup

Hop311:
- Changed locale CSVs to use `;` delimeters
- Added missing localisation keys (only thing missing now should be refresh rate options)
- Added `SettingCheckBox` class for "Explode Ears" option and made it control startup music
- Set lots of UI elements to `FOCUS_NONE`


game/src/Game/Menu/OptionMenu/MonitorDisplaySelector.gd
```diff
@@ -3,9 +3,18 @@ extends SettingOptionButton
 func _setup_button() -> void:
 	clear()
 	for screen_index in DisplayServer.get_screen_count():
-		add_item("Monitor %d" % (screen_index + 1))
+		add_item(str(screen_index + 1))
```
Spartan322, line 6: Wouldn't this produce different results to https://github.com/OpenVicProject/OpenVic/blob/75bb77a570198d1c9300018aba1975b42fa7f7c8/game/src/Game/Menu/OptionMenu/MonitorDisplaySelector.gd#L16

Wouldn't this mean that in the default translation, without the player changing it back and forth, it doesn't attempt any translation?

Also you might want to use [`TextServer.format_number(String number, String language = "")`](https://docs.godotengine.org/en/stable/classes/class_textserver.html#class-textserver-method-format-number) for the number's translation.

Spartan322 (changes requested):
Almost everything seems good, just think my one thing in MonitorDisplaySelector should be addressed or explained.

game/src/Game/Menu/OptionMenu/MonitorDisplaySelector.gd
Hop311: `(str(screen_index + 1)` is a placeholder value that should always be overridden by `_update_monitor_options_text`, I changed it to see if `NOTIFICATION_TRANSLATION_CHANGED` was emitted after creation. I think it's best to have the localisation code in only one place and to add comments making clear that `add_item` is being given placeholder values, eliminating ambiguity around what code actually controls the display text.

Spartan322: Does the placeholder actually get replaced after `_setup_button` is called?

Hop311: The placeholder is getting replaced from what I saw, because of a translation notification, I relied on that thinking it better to avoid unnecessary consecutive updates but thinking again the notification can’t be relied on unless it’s is guaranteed on creation - I suspect it might instead be coming from the locale button created with option menu. I’ll foolproof the placeholder updating code and add in the better number formatting tomorrow.

Spartan322 (approved):

PR #132, opened 2023-06-12
Basic Loading Screen

BrickPi:
Adds a basic loading screen, refactors loading to take place on a background thread after the loading screen is rendered, so that the loading screen UI can inform the player of progress. Feel free to check it out and air any suggestions.


game/src/Game/LoadingScreen.gd
```diff
@@ -0,0 +1,32 @@
...
+
+@export var progress_bar: ProgressBar
+@export var quote_label: Label
+
+var loadthread: Thread
+var quotes: PackedStringArray = []
+
+func update_loading_screen(percent_complete: int, quote_should_change = false):
+	if percent_complete > 99:
+		get_tree().change_scene_to_file("res://src/Game/GameMenu.tscn")
```
Hop311, line 11: I think this should be moved to the end of `load_events`, rather than relying on actually hitting 100%, especially if the reported percentages become dynamic rather than fixed how they are now.

game/src/Game/Autoload/Events.gd
```diff
@@ -19,9 +19,10 @@ var _define_filepaths_dict : Dictionary = {
 
 # REQUIREMENTS
 # * FS-333, FS-334, FS-335, FS-341
-func _ready():
+func load_events():
```
Hop311, line 22:
```suggestion
func load_events(loading_screen : LoadingScreen):
```
Better to avoid hardcoded nodepaths, replace `$/root/GameStartup/LoadingScreen` with `loading_screen` below, maybe even make them `if loading_screen: loading_screen.update_loading_screen(...)` so that loading could still happen without the screen.

game/src/Game/LoadingScreen.gd
```diff
@@ -0,0 +1,32 @@
...
+	if quote_should_change:
+		quote_label.text = quotes[randi() % quotes.size()]
+
+func _on_splash_container_splash_end():
+	show()
+
+func _ready():
+	# FS-3, UI-30, UIFUN-35
+	loadthread = Thread.new()
+	loadthread.start(Events.load_events)
```
Hop311, line 23:
```suggestion
	loadthread.start(Events.load_events.bind(self))
```
Given the thread the `LoadingScreen` that queued it, instead of using the hardcoded nodepath in `load_events`.

Hop311 (approved):
Great work!

game/src/Game/Autoload/Events.gd
```diff
@@ -33,13 +34,26 @@ func _ready():
...
-
+	
+	GameDebug = load("res://src/Game/Autoload/Events/GameDebug.gd").new()
+	loading_screen.update_loading_screen(15)
+	Options = load("res://src/Game/Autoload/Events/Options.gd").new()
+	loading_screen.update_loading_screen(25)
+	Localisation = load("res://src/Game/Autoload/Events/Localisation.gd").new()
+	loading_screen.update_loading_screen(45)
+	ShaderManager = load("res://src/Game/Autoload/Events/ShaderManager.gd").new()
+	loading_screen.update_loading_screen(50, true)
```
Spartan322, line 45:
```suggestion
	GameDebug = preload("res://src/Game/Autoload/Events/GameDebug.gd").new()
	loading_screen.update_loading_screen(15)
	Options = preload("res://src/Game/Autoload/Events/Options.gd").new()
	loading_screen.update_loading_screen(25)
	Localisation = preload("res://src/Game/Autoload/Events/Localisation.gd").new()
	loading_screen.update_loading_screen(45)
	ShaderManager = preload("res://src/Game/Autoload/Events/ShaderManager.gd").new()
	loading_screen.update_loading_screen(50, true)
```

Since they're hardcoded right now, `preload` is preferable to `load`.

game/src/Game/LoadingScreen.gd
```diff
@@ -0,0 +1,30 @@
...
+	if quote_should_change:
+		quote_label.text = quotes[randi() % quotes.size()]
+
+func _on_splash_container_splash_end():
+	show()
+
+func _ready():
+	# FS-3, UI-30, UIFUN-35
+	loadthread = Thread.new()
+	loadthread.start(Events.load_events.bind(self))
```
Spartan322, line 21: Probably best to consider [WorkerThreadPool](https://docs.godotengine.org/en/latest/classes/class_workerthreadpool.html) (yes this documentation is 4.1+, but the API should not change, the backend is better in 4.1, but how you use it should not change regardless) whether now or later.

Spartan322 (changes requested):
Mostly decent, only the load to preload I think should be changed, could use more requirement comments, whether the threading in LoadingScreen is changed or not can be a later consideration, just making comment of it here for consideration.

Spartan322 (approved):

PR #140, opened 2023-07-17
Update quotes.txt

FarmingtonS9:
FS-3
Added quotes from "Loading Quotes" spreadsheet.


FarmingtonS9:
In the next version, we should add more quotes with an economic theme.

Hop311 (approved):
LGTM, will eventually need to be localised somehow but forr now these are great

PR #141, opened 2023-07-17
Update to Godot 4.1.1 Stable Release

Spartan322:
Add bash-based Godot URL environment variable handling to openvic-env action
Separate Commit environment variable handling in openvic-env to self-contained js file.

If `GODOT_BASE_DOWNLOAD_URL`, `GODOT_VERSION`, and `GODOT_VERSION_TYPE` are all set, this will then set a `GODOT_LINUX_URL` and `GODOT_TEMPLATE_URL` environment variable that points to the proper download path for the template and x86_64 linux version of the respective binaries.

`GODOT_BASE_DOWNLOAD_URL` must be either `downloads.tuxfamily.org/godotengine` or `github.com/godotengine` (must point to a repo, accounts for also pointing to releases or releases/download) for now.
Errors if either `GODOT_LINUX_URL` or `GODOT_TEMPLATE_URL` is not set.
If `GODOT_LINUX_URL` environment variable is already set to a non-empty value, it does not modify it. If it does not exist as a valid link, it does not set `GODOT_LINUX_URL`.
If `GODOT_TEMPLATE_URL` environment variable is already set to a non-empty value, it does not modify it. If it does not exist as a valid link, it does not set `GODOT_TEMPLATE_URL`.

### DIFF
--- a/.github/actions/openvic-env/action.yml
+++ b/.github/actions/openvic-env/action.yml
@@ -1,33 +1,13 @@
 name: Setup OpenVic Environment
 description: Setup OpenVic Environment
 runs:
-    using: "composite"
-    steps:
-        - name: Setup Environment
-          uses: actions/github-script@v6
-          with:
-            script: |
-                const commit_name = 'OPENVIC_COMMIT';
-                const tag_name = 'OPENVIC_TAG';
-                const release_name = 'OPENVIC_RELEASE';
-                var commit_sha = process.env["GITHUB_SHA"];
-                if(context.eventName === 'pull_request') {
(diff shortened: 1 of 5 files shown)


Hop311 (approved):
Godot 4.1.1 working for me, and the Bash script looks good, I presume it's working since the checks passed.

Spartan322:
The whitespace is a bit off, let me improve this a little, I'm gonna do is put the two environment setup scripts into their own files instead.

Hop311 (approved):

PR #144, opened 2023-07-27
First pass at pops (test pops and density mapmode)

Hop311:
- Update sim submodule to bring in Pop/Culture/Religion data structures (currently populated with hardcoded test instances)
- Added population density mapmode

Requirements implemented:
- MAP-33, UI-121, UI-562, UIFUN-141

### DIFF
--- a/extension/src/GameSingleton.cpp
+++ b/extension/src/GameSingleton.cpp
@@ -1,5 +1,7 @@
 #include "GameSingleton.hpp"
 
+#include <cassert>
+
 #include <godot_cpp/variant/utility_functions.hpp>
 
 #include "openvic/utility/Logger.hpp"
@@ -8,6 +10,16 @@
 
 using namespace godot;
 using namespace OpenVic;
+
+TerrainVariant::TerrainVariant(std::string const& new_identfier,
+	colour_t new_colour, Ref<Image> const& new_image)
+	: HasIdentifier { new_identfier },
+	  HasColour { new_colour, true },
+	  image { new_image } {}
+
+Ref<Image> TerrainVariant::get_image() const {
+	return image;
+}
 
 GameSingleton* GameSingleton::singleton = nullptr;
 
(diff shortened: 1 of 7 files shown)


extension/src/GameSingleton.cpp
```diff
@@ -9,6 +11,18 @@
 using namespace godot;
 using namespace OpenVic;
 
+TerrainVariant::TerrainVariant(std::string const& new_identfier,
+	colour_t new_colour, Ref<Image> const& new_image)
+	: HasIdentifier { new_identfier },
+	  HasColour { new_colour, true },
+	  image { new_image } {
+	assert(image.is_valid());
```
Spartan322, line 19: Probably should be using Godot erroring instead of asserts.

Spartan322 (changes requested):

extension/src/GameSingleton.cpp
Hop311: The actual null-checking is the responsibility of whoever calls the constructor (since constructors can't an return error value), if you look up `terrain_variants.add_item` there are 3 places it's used all with their own checks and custom error messages. The `assert` here and in similar constructors like `HasIdentifier` and `HasColour` are just to remind me of the conditions and catch silly mistakes. Still, the constructor shouldn't be publicly accessible, I'll re-commit with it private and `GameSingleton` added as a friend class.

Spartan322: Well if you're gonna runtime asserts at all, you should be using Godot's erroring macros else if you miss a case it just crashes the entire engine.

Hop311: Re-committed without the assert, `TerrainVariant` itself would still be fine with an invalid reference, it would only become an issue when the terrain image stitching happens which is right next to the code which already does the actual invalid reference check.

Spartan322 (approved):

PR #146, opened 2023-07-29
Add Piecharts (gd 4.1)

Nemrav:
- Adds Pie/Donut charts, this time as godot 4.1 scenes (upgrade broke previous pr)
- Adds placeholder piechart to the province overview panel

Pie charts related scripts and nodes are located at src/game/theme/piechart

Areas of concern/known issues:
- Resources are wasted as a "new" shader is instanced for every pie chart. This is because, for now, "instance uniforms" aren't supported for 2d shaders.
- To set the parameters of the shader, there is something of variable shadowing/passthrough from the piechart.gd for convenience's sake, due to the above lack of instance uniforms on the chart shader.
- The tooltip does not always go away when the mouse exits the area. I don't know how to fix this. Custom tooltips are used here to allow for rich text (ie. bolding and underlining the text).

Future improvements:
- I expect requests for code style improvements
- looks to be tweaked some more
- Additions to the RichTooltips


Hop311:
As well as Nemrav's great work on pie charts, this PR includes:

# Main repo
- Piecharts are generated at the godot-cpp level (using Nemrav's shader translated into Godot C++) and then cached until next updated. In the future this should be changed to use the overlay textures (and ideally the Godot draw functions)
- Added province pop type, ideology and culture pie charts, hooked up to receive and display data from the simulation (although ideology doesn't send anything yet because political stuff hasn't been implemented)
- Added culture/nationality mapmode, for now only showing largest culture per province (this PR doesn't include culture loading, that'll be in the next one)
- Removed some chained signals for load progress starting, changing, and finishing, as they already have global signals in `Events`

# Sim submodule
- Created `HasIdentifierAndColour` (inheriting from both `HasIdentifier` and `HasColour`), for use in `distribution_t = std::unordered_map<HasIdentifierAndColour const*, float>`, the C++-level representation of piechart data. (Might be better to use `std::map`?). Currently provinces generate `distribution_t`s for pop type and culture,
- Replaced a lot of `std::string const&` args with `const std::string_view` to reduce unnecessary string copies, including having `IdentifierRegistry::get_item_by_identifier` search for a `std::string_view` in a `std::string`-keyed `std::map` without needing to create a temporary `std::string`
- Clear pops before generating test pops in `Map::setup`, so that resigning to the main menu and then going back into the game doesn't add on to the pops from the previous session
- `Date::from_string` rewritten to take `std::string_view` argument and `std::from_chars` to turn parse ints (previously we checked the string was valid then passed it to `stoi`, hoping it wouldn't throw an exception we hadn't considered). Once Fakebyte's fixed point stuff is merged we could use its `string_to_int64` function, if there's any advantage over `std::from_chars`.
- Rearranged province building generation to make sure the province's building `IdentifierRegistry` locks even after encountering an error
- Added error message for invalid input to `Map::set_selected_province`
- Added summary of province and terrain image loading errors (e.g. unrecognised colours or missing provinces) and the option to disable the full errors (very spammy with vic2's province image which contains many mistake pixels and doesn't contain a fair few provinces defined in `definition.csv`)
- Added check + error message to prevent adding a Pop to a water province
- Moved province pop stats updating from a their own individual functions to `Province::update_pops`, since they'll never be updated in isolation and this way we only need to iterate over the province's pops once per update
- Fleshed out `ProvinceSet` with its own `add_province`, `lock`, `is_locked` and `reset` functions, similar to `IdentifierRegistry`,  updated `Region` generation to use the new methods and changed `Map` to use a `ProvinceSet` to store all water provinces (while still setting a flag in the provinces themselves)
- Reordered the arguments of various constructors and constructing functions to match the order the variables are actually initialised in (mostly moving `identifier` and `colour` to the front)

Spartan322 (approved):
Approved for now, but I really want to refurbish everything going on here, move things to proper places, use the proper functions, organize behavior out of Utilities and GameSingleton.

PR #147, opened 2023-08-07
Added religion.json data

FarmingtonS9:
Added religion data in json format.

Working on requirements:
- POP-286 to POP-299

### DIFF
--- a/game/common/religion.json
+++ b/game/common/religion.json
@@ -96,8 +96,8 @@
             ]
         }
     },
-    "animist": {
-        "pagan": {
+    "pagan": {
+        "animist": {
             "colour": [
                 0.5,
                 0,
--- a/game/common/religion.json
+++ b/game/common/religion.json
@@ -1,0 +1,108 @@
+{
+    "christian": {
+        "catholic": {
+            "colour": [
+                0.8,
+                0.8,
+                0
+            ]
+        },
+        "protestant": {
+            "colour": [
+                0,
+                0,
+                0.7
+            ]
+        },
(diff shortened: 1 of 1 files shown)


game/common/religion.json
```diff
@@ -0,0 +1,108 @@
...
+        "sikh": {
+            "colour": [
+                0.3,
+                0.8,
+                0
+            ]
+        }
+    },
+    "animist": {
+        "pagan": {
```
Spartan322, line 100: Shouldn't this be the opposite, all animists are pagans, but not all pagans are animist, and animist is what the game shows as the religion.

Spartan322 (changes requested):

game/common/religion.json
FarmingtonS9: True, fixing it now

Spartan322 (changes requested):
You should remove the submodule push, it is 4 commits behind the master, that's why it won't compile, remember that when submodules change you need to run ` git submodule update --init --recursive` to update your submodule, and don't just stage every change.

Spartan322:
Also when you're ready, you should squash this, an easy way to do this is `git rebase -i HEAD~2` 

The 2 here can be replaced with however many (non-merge) commits you have and replace every `pick` but the first one with `s` or `squash`.

FarmingtonS9:
Yea, still new to Git commands. I hadn't realised that I hadn't updated the submodules.

Hop311:
Closing as we won't be implementing custom mode for a while.

PR #148, opened 2023-08-10
Base Fixed Point class and Sine LUT

FakeByte:
-Not conforming with our style guide yet
-Probably wont compile since the include paths need to be fixed after I copy and pasted it into this repo
-Math class only supports sine for now, more added later


Hop311:
Closing as this has been moved to the OpenVic-Simulation submodule.

PR #150, opened 2023-08-24
Add signals to MusicConductor

Spartan322:
Enable music on startup
Add sound option loading on startup
Fix MusicPlayer play button starting with incorrect text


Hop311 (approved):
LGTM 👍

PR #151, opened 2023-08-25
Build system cleanup + dataloading scaffolding + JSON dataloading removed

Hop311:
- Moved `openvic-dataloader` submodule to `openvic-simulation`.
- Modified `SConstruct` to build `openvic-simulation` as a library using its own `SConstruct`, similar to what `openvic-simulation` does to `openvic-dataloader` and `openvic-dataloader` does to `lexy`.
- Moved extension sources into `openvic-extension` subdir.
- Removed JSON based "custom mode" loading (compatibility mode is the current focus).
- Added `Logger::warning` (alongside `Logger::info` and `Logger::error`).
- Moved "hardcoded defines" (mapmodes, building types) to simulation submodule.
- Added religion mapmode.
- Added prototype gdscript-level `AssetManager` which queries the C++ layer to lookup asset-paths and then loads and caches them, making them available to UI nodes.
- Added methods to pass a list of root directories to the C++ layer, those being the base install directory + an ordered list of mod directories. Currently this has no UI, the base directory must be passed as a command line argument after `--compatibility-mode` or edited in `GameStart.gd`, and any mod directories also need to be edited in `GameStart.gd`.

Requirement commented:
- MAP-34
- UI-24, UI-124, UI-125, UI-127, UI-563
- UIFUN-142


BrickPi (approved):
works as expected, tested on Win64; code lgtm.

PR #152, opened 2023-08-27
Update README.md for Archlinux users

Rerum02:
Instructions are made to be easy to follow so that one can install and set up OpenVic, for Arch based distribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,8 @@ For detailed instructions, view the Contributor Quickstart Guide [here](docs/con
 * [scons](https://scons.org/)
 
 ## Repo Setup
-1. Clone the OpenVic Repo to a suitable folder using the git command "git clone https://github.com/OpenVicProject/OpenVic.git"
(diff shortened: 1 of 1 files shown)


Spartan322:
Outside of repo setup quotation being fixed, I honestly don't think this is warranted, this would then require us to add Windows, Mac, and Ubuntu install instructions with the same detail, besides Arch users don't all carry yay and both of the packages are kinda the responsibility of the platform user, its get messy trying to figure that out for every user. I suppose I might not mind if this were somewhere else other then the readme, but even so it should require equal treatment of all the biggest platforms.

Rerum02:
That's understandable, I should be able to add Windows, Deb (which would include Ubuntu, mint, ect), and Fedora  as well with the same amount of detail, as for Mac that'll take me a little extra time. As for using yay, I can just add installation for yay at the beginning, it would only add one extra line.

Spartan322:
Even in all this, I think the best thing would be to put detail instructions in its own separate documentation and link the readme to it instead.

PR #153, opened 2023-08-28
Add Window Override

Spartan322:
Hides window for help argument (`./Openvic -- --help`)
Hides window until video options loading is complete

Makes Window handling way more convenient, can delay a visible window until after everything has been processed and prepared to display it.
Needs Windows and Mac testing.


Hop311 (approved):
Works for me

PR #160, opened 2023-09-27
Terrain types

Hop311:
- Added GLI submodule + runtime DDS loading function (https://github.com/godotengine/godot-proposals/issues/5748).
- Added `GameSingleton.load_image` to enable runtime DDS loading from GDScript
- Moved terrain and province BMP loading to sim submodule, removing `TerrainVariant` struct and moving what's left of `LoadGameCompatibility.cpp` and `LoadGameOpenVic.cpp`
- Added terrain type province info key and display label.


SConstruct
```diff
@@ -36,9 +36,10 @@ SConscript("extension/deps/SCsub", "env")
 # - LINKFLAGS are for linking flags
 
 # tweak this if you want to use different folders, or more folders, to store your source code in.
-paths = ["extension/src"]
-env.Append(CPPPATH=[[env.Dir(p) for p in paths]])
-sources = env.GlobRecursive("*.cpp", paths)
+source_paths = ["extension/src"]
+include_paths = ["extension/deps/gli", "extension/deps/gli/external"]
```
Spartan322, line 40: This should be added in the `extension/deps/SCsub`, not directly in the SConstruct, adding deps should always rely on `extension/deps/SCsub` so we know where dependencies are defined for consistently, only godot-cpp is getting special treatment over this.

All that needs to be added to the SCsub is:
```py
def build_gli(env):
	gli_includes = ["gli", "gli/external"]
	env.gli_loader = {}
	env.gli_loader["INCPATH"] = [env.Dir(p) for p in gli_includes]
	env.Append(CPPPATH=env.gli_loader["INCPATH"])

build_openvic_simulation(env)
build_gli(env)
```

Spartan322 (changes requested):

Spartan322 (dismissed):

Spartan322 (approved):

PR #163, opened 2023-11-10
Add ClassBinding helpers

Spartan322:
depends on OpenVicProject/scripts#5

Adds OV_BIND_METHOD
Adds OV_BIND_SMETHOD
Adds OV_BIND_SMETHOD_L

Change _bind_methods to use ClassBinding helpers
Add utility/StringLiteral
Make `GameSingleton::singleton` inline
Move `OpenVic::Checksum` implementation to source file

Update scripts to [ce1aef8d7d9d5ba9851a1abdd981d3b796024079](https://github.com/OpenVicProject/scripts/commit/ce1aef8d7d9d5ba9851a1abdd981d3b796024079)

### DIFF
--- a/extension/src/openvic-extension/GameSingleton.cpp
+++ b/extension/src/openvic-extension/GameSingleton.cpp
@@ -1,11 +1,14 @@
 #include "GameSingleton.hpp"
 
+#include <godot_cpp/core/class_db.hpp>
+#include <godot_cpp/core/error_macros.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
 #include <openvic-simulation/utility/Logger.hpp>
 
 #include "openvic-extension/LoadLocalisation.hpp"
 #include "openvic-extension/Utilities.hpp"
+#include "openvic-extension/utility/ClassBindings.hpp"
 
 using namespace godot;
 using namespace OpenVic;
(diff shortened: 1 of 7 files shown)


extension/src/openvic-extension/utility/StringLiteral.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+namespace OpenVic {
+	template<std::size_t N>
+	struct StringLiteral {
+		constexpr StringLiteral(const char (&str)[N]) {
+			std::copy_n(str, N, value);
+		}
+
+		constexpr StringLiteral(std::string_view string) {
+			std::copy_n(string.begin(), N, value);
```
schombert, line 17: This will happily copy from data after the end of the string view, which may fault, depending where the string view is pointing

extension/src/openvic-extension/utility/StringLiteral.hpp
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		char value[N];
+		static const constexpr std::integral_constant<std::size_t, N - 1> size {};
+
+		struct iterator {
```
schombert, line 23: Strictly speaking this struct is unnecessary. char* satisfies the requirements of an iterator, and so begin and end could just return the relevant pointers, given that this iterator doesn't provide any functionality beyond that of char*

extension/src/openvic-extension/utility/StringLiteral.hpp
```diff
@@ -0,0 +1,104 @@
...
+		private:
+			pointer _ptr;
+		};
+
+		constexpr iterator begin() const {
+			return iterator { &value };
+		}
+
+		constexpr iterator end() const {
+			return iterator { &value + N };
```
schombert, line 93: Since you have a string view constructor, this could potentially be constructed with a string shorter than N characters, in which case, you probably don't want to iterate past the end of the actual data

extension/src/openvic-extension/utility/StringLiteral.hpp
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		constexpr operator std::string_view() const {
+			return std::string_view { value, N };
```
schombert, line 97: as above

extension/src/openvic-extension/utility/StringLiteral.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+		char value[N];
+		static const constexpr std::integral_constant<std::size_t, N - 1> size {};
+
+		struct iterator {
+			using iterator_concept [[maybe_unused]] = std::contiguous_iterator_tag;
+			using difference_type = std::ptrdiff_t;
+			using element_type = char; // element_type is a reserved name that must be used in the definition
+			using pointer = element_type*;
+			using reference = element_type&;
```
schombert, line 28: I note also that this is a non-constant iterator (you could modify the data pointed to using it), but you return it from functions marked as const. Thus you are creating a potential const "leak"

extension/src/openvic-extension/utility/StringLiteral.hpp
Spartan322: Should be fixed running as consteval and having an assert that ensures `string.size() == N`

extension/src/openvic-extension/utility/StringLiteral.hpp
Spartan322: With the consteval constructor, should no longer be the case.

extension/src/openvic-extension/utility/StringLiteral.hpp
Spartan322: With the consteval constructor, should no longer be the case.

extension/src/openvic-extension/GameSingleton.cpp
```diff
@@ -14,94 +17,81 @@ using OpenVic::Utilities::godot_to_std_string;
...
+	OV_BIND_METHOD(GameSingleton::toggle_paused);
+	OV_BIND_METHOD(GameSingleton::is_paused);
+	OV_BIND_METHOD(GameSingleton::increase_speed);
+	OV_BIND_METHOD(GameSingleton::decrease_speed);
+	OV_BIND_METHOD(GameSingleton::can_increase_speed);
+	OV_BIND_METHOD(GameSingleton::can_decrease_speed);
+	OV_BIND_METHOD(GameSingleton::get_longform_date);
+	OV_BIND_METHOD(GameSingleton::try_tick);
+	OV_BIND_METHOD(GameSingleton::get_mapmode_count);
+	OV_BIND_METHOD(GameSingleton::get_mapmode_count);
```
Hop311, line 58: Duplicates of L40, these both need to be removed

extension/src/openvic-extension/GameSingleton.cpp
```diff
@@ -14,94 +17,81 @@ using OpenVic::Utilities::godot_to_std_string;
...
+	OV_BIND_SMETHOD(get_building_info_expansion_state_key);
+	OV_BIND_SMETHOD(get_building_info_start_date_key);
+	OV_BIND_SMETHOD(get_building_info_end_date_key);
+	OV_BIND_SMETHOD(get_building_info_expansion_progress_key);
+
+	OV_BIND_SMETHOD(get_piechart_info_size_key);
+	OV_BIND_SMETHOD(get_piechart_info_colour_key);
+
+	OV_BIND_SMETHOD(get_province_info_province_key);
+	OV_BIND_SMETHOD(get_province_info_province_key);
```
Hop311, line 85: Duplicates of L63, these both need to be removed

Hop311 (approved):

PR #166, opened 2023-11-13
GUI elements -> Godot UI nodes generator

Hop311:
- `AssetManager` singleton for loading images and fonts:
    - For `Image`s, `get_image` takes in a defines-relative `StringName` path, looks it up using `GameSingleton`'s `Dataloader::lookup_image_file_or_dds`, and loads it. `Image`s are cached with the defines-relative `StringName` path as a key, in case they're needed again.
    - For `Texture`s, `get_texture` takes in a defines-relative `StringName` path, loads it as an `Image` as described above, and uses it to generated an `ImageTexture`. `Texture`s are also cached with the defines-relative `StringName` path as a key, so it's always preferrable to use `get_texture` over making an `ImageTexture` from a `get_image`-sourced `Image` yourself.
    - For `Font`'s, `get_font` takes just the name of the font as a `StringName`, with no folder or file extension, and then looks up and loads the image `gfx/font/font_name.tga` and Windows font definition file `gfx/font/font_name.fnt`. Once loaded for the first time, `Font`s are cached for reuse with their `StringName` name as a key.
- Basic/prototype system for generating Godot UI nodes from GUI elements (UIAdapter files):
    - `GameSingleton.generate_gui(gui_file, gui_element)` function, callable from GDScript, takes a `.gui` file name and the name of a top-level GUI element in that file, e.g. "province_interface.gui" and "province_view", and returns a `Control`-derived node generated from that element, including its recursively-generated child elements.
    - With the current setup, the `.gui` file needs to be in a list in `Dataloader::_load_interface_files` in the SIM submodule so that's it's loaded with the rest of the game's defines; we could change it to load them when they're requested from GDScript, but that could cause unwanted lag spikes.
    - I plan to replace/adapt this sytem to instead have a custom Godot Node class which you add to a scene in the editor, set its `.gui` file and element name properties and then have it automatically generate. This class will also have helper functions for accessing and modifying child elements, e.g. the ability to pass it a list of `(node_path, expected_node_type)` pairs to be looked up, type checked and returned, rather than doing those steps manually from GDScript for each child element you want to access/modify.
- Changed `ProvinceOverviewPanel` to use the new GUI generation system, with it's behaviour control script modified to work with the new layout. All the static `StringName`-returning `GameSingleton.get_***_info_***_key()` functions have been removed and replaced with simple constants in the C++ and GDScript code where they were used, the mess they created wasn't worth the benefit of preventing mistakes in the key `StringName`s.
- Rearranged files into `classes` (based off of `godot-cpp` file structure) and `singletons` folders.
- Added `Utilities::std_to_godot_string_name` and `Utilities::std_view_to_godot_string_name` to avoid needing an intermediate `godot::String`.

- Added support for striped province colouring, currently used by culture and religion mapmodes.
- Used land and water colormap textures to tint the cosmetic terrain map.


BrickPi (approved):
Very important stuff, looks good. I might see about extending that other documentation PR.

BrickPi (approved):

PR #177, opened 2023-12-23
Add constant GameMenuScene preload to GameStart.gd

Spartan322:
Add type hints to GameStart.gd constants
Seperate section_name and setting_name from Nodes group in inspector
Prefer callable's called_deferred to string caller in GameStart.gd
Change `GameStart._initialize_game` to use change_scene_to_packed and GameMenuScene constant to change scene instead

This seems to result in a quicker transition from loading screen to main menu.


Hop311 (approved):
Tested on Windows, noticably faster transition, good work 👍

PR #178, opened 2023-12-23
Apply type hints to menu scripts

Spartan322:


game/src/Game/Menu/OptionMenu/ResolutionSelector.gd
```diff
@@ -66,15 +66,15 @@ func _setup_button() -> void:
 	else:
 		_sync_resolutions()
 
-func _get_value_for_file(select_value : int) -> Variant:
+func _get_value_for_file(select_value : int):
```
Hop311, line 69: Why remove the return type here?

Hop311 (approved):

game/src/Game/Menu/OptionMenu/ResolutionSelector.gd
Spartan322: It was causing strange inheritance errors saying that `_get_value_for_file` doesn't inherit the signature of a base class. (despite it doing just that)

PR #185, opened 2023-12-29
Document RGO & social income

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -36,9 +32,34 @@ POPs:
...
+    Everyday & luxury needs scalar: Life needs scalar * (1 + number of inventions * defines.pops.INVENTION_IMPACT_ON_DEMAND)
+  POP income:
+    Pensions: 2 * pension_level * administrative efficiency * social spending slider * base life needs price * (1 - effective tax) * POP size / 200000
+    Unemployment subsidies: 2 * unemployment_benefit * administrative efficiency * social spending slider * base life needs price * (1 - effective tax) * POP size / 200000
+    Wages:
+      Minimum (worker) wage: 2 * minimum_wage * administrative efficiency * base life needs price * (1 - effective tax) * POP size / 200000
+      RGO wages:
+        Owners:
+          if minimum worker wages > normal worker wage:
+            formula: (RGO income - total worker income) * (1 - effective tax)
```
Hop311, line 48: Shouldn't this include a `owner POP size / owners` term, like for the `else` case below and for both worker cases, so that `RGO income - total worker income` is divided between owner POPs proportional to their size? Otherwise it'll create money out of thin air based on the number of distinct owner POPs, e.g. if a single POP of 1000 owners gets paid `£ x`, then the same population split into two POPs of size 500 would get paid a total of `£ 2x`.

docs/simulation/calculations.yaml
```diff
@@ -36,9 +32,34 @@ POPs:
...
+            formula: RGO income * min(0.5, 2 * owners / workers) * owner POP size / owners * (1 - effective tax) 
+            owners: total size of owner POPs in state
+            workers: number of employed workers in RGO
+            min(0.5: Hardcoded maximum of half the RGO income.
+            2*: Hardcoded value.
+        Workers:
+          if is_slave: 0
+          else:
+            if minimum wage > normal wage:
+              formula: max(minimum wage, RGO income * employed workers in POP / total non-slave employed workers) * (1 - effective tax)
```
Hop311, line 59: If there are no slaves workers, since each worker pop gets a share of the total `RGO income` proportional to their size (or more if `minimum wage` is greater), won't `total worker income` on line 48 always be `>= RGO income`, and so owner wages will always be `<= 0` when `minimum worker wages > normal worker wage`? (This is still the case with a `owner POP size / owners` added like I mention in my comment on line 48.)

docs/simulation/calculations.yaml
```diff
@@ -36,9 +32,34 @@ POPs:
...
+          else:
+            formula: RGO income * min(0.5, 2 * owners / workers) * owner POP size / owners * (1 - effective tax) 
+            owners: total size of owner POPs in state
+            workers: number of employed workers in RGO
+            min(0.5: Hardcoded maximum of half the RGO income.
+            2*: Hardcoded value.
+        Workers:
+          if is_slave: 0
+          else:
+            if minimum wage > normal wage:
```
Hop311, line 58: Just to make sure I understand correctly, here (and above on line 47) `minimum wage` is the `Minimum (worker) wage` calculated on line `44`, and `normal wage` the amount calculated after the `else` on line 61? So in practice we'll calculate the `Minimum (worker) wage` and the `normal worker wage`, and if the minimum wage is bigger then we calculate and use the formulae on lines 48 and 59, otherwise we just use the already-calculated `normal worker wage` and calculate and use the formula on line 50?

Hop311 (commented):
Great work, exactly what we need!
I may be wrong, but I think there are some missing terms in the `RGO wages` formulae, pretty sure of it on line 48, and semi-sure on line 59 but that could just be paradox using silly formulae.

docs/simulation/calculations.yaml
wvpm: You're right. I forgot to add that.
Background: I copied some old notes and edited them slightly. I missed this line.

docs/simulation/calculations.yaml
wvpm: Minimum wage does not depend on RGO income.
Normal worker wage is a fraction of the RGO income.

"won't total worker income on line 48 always be >= RGO income [?]"
No.
If the RGO is profitable, the normal wage can be greater than the minimum wage.
Also total worker wage is capped at RGO income.
Even if minimum wage > normal wage, there can be plenty of profit left for the owners.

Example (no taxes):
RGO income = 200
employed workers (1 pop) = 10000
owners in state = 100

Normal total worker wage = 200*(1-2*100/10000) = 196
So per individual = 0.0196
Normal total owner wage = 200*2*100/10000 = 4

If minimum wage per individual ended up being 0.0198
Then each worker would get that and
Total owner wage = (200 - 10000 * 0.0198) = 2

So even with minimum wage, the owners can have a wage > 0.

"and so owner wages will always be <= 0"
Owner wages can never go negative.

docs/simulation/calculations.yaml
wvpm: "minimum wage is the Minimum (worker) wage calculated on line 44"
Yes. Owners don't have minimum wage, that's why I added "(worker)".

"and normal wage the amount calculated after the else on line 61"
Yes. I struggled documenting this...

"So in practice we'll calculate the Minimum (worker) wage and the normal worker wage [...]"
Yes. Note that minimum wage differs **per country** and **pop type**, not per RGO, province, POP or anything else.

docs/simulation/calculations.yaml
Hop311: My concern was the second argument in the `max` function:
```
max(minimum wage, RGO income * employed workers in POP / total non-slave employed workers)
```
If we're in the one worker scenario, `employed workers in POP / total non-slave employed workers = 1`, so this becomes:
```
max(minimum wage, RGO income)
``
If `RGO income` is greater than minimum wage, then the worker gets the whole `RGO income` as its wages, or am I missing something?

wvpm: Thanks for pointing that out. The max should be min. So either minimum wage or a proportional share of the RGO income. This is what I meant with "Also total worker wage is capped at RGO income.".

Hop311: Ah ok, so it's still possible for an owner to get 0 wages, but only if minimum wages are so high that there's non left for him, which makes sense.

Hop311 (approved):

PR #186, opened 2023-12-31
Document influence modifiers

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -15,10 +15,38 @@ Budget:
...
+    Base daily influence gain per country:
       if total priority bars > 0: Total base influence gain * priority bars / total priority bars
       else: Total base influence gain / number of countries (excluding banned)
+    Country influence gain modifier:
+      note: 'our' and 'we' refers to the influencer. 'target' is the country being influenced.
+      formula: 1 + score mod + relations mod + population mod + investment mod + puppet mod + neighbour mod + continent mod + discredit mod
+      score mod: 1 - target score / our score
+      relations mod: relations / defines.diplomacy.RELATION_INFLUENCE_MODIFIER
+      population mod:
+        formula: truncate((target population - LARGE_POPULATION_LIMIT) / LARGE_POPULATION_INFLUENCE_PENALTY_CHUNK) * LARGE_POPULATION_INFLUENCE_PENALTY
```
Hop311, line 28: Is `truncate` here downwards or towards zero? Either way `1.23` will go to `1`, but will `-1.23` go to `-2` (downwards) or `-1` (towards zero)?

Hop311 (approved):
Great work!

docs/simulation/calculations.yaml
wvpm: I used the word truncate as I'm used to it just removing the fractions. Ie rounding towards 0.
Since all 3 values are always positive, I'll just call it floor.

Hop311: The values are positive but the result of the equation can be negative. Using the vanilla defines values for a target country with a population of 1,000,000:
```
target population = 1,000,000
LARGE_POPULATION_LIMIT = 9,000,000
LARGE_POPULATION_INFLUENCE_PENALTY_CHUNK = 250,000

(target population - LARGE_POPULATION_LIMIT) / LARGE_POPULATION_INFLUENCE_PENALTY_CHUNK
= (1,000,000 - 9,000,000) / 250,000 = -8,000,000 / 250,000 = -32
```
However I believe this modifier only applies for target populations greater than `LARGE_POPULATION_LIMIT`, so the equation either needs a `max(0, ...)` or `if target population > LARGE_POPULATION_LIMIT ... else 0`.

Hop311 (approved):

PR #187, opened 2024-01-08
Updated SIM submodule and switched to ordered_map

Hop311:


extension/src/openvic-extension/singletons/AssetManager.cpp
```diff
@@ -45,34 +45,34 @@ Ref<Image> AssetManager::_load_image(StringName const& path) {
 	return image;
 }
 
-AssetManager::image_asset_map_t::iterator AssetManager::_get_image_asset(StringName const& path) {
-	const image_asset_map_t::iterator it = image_assets.find(path);
+AssetManager::image_asset_t* AssetManager::_get_image_asset(StringName const& path) {
+	image_asset_map_t::iterator it = image_assets.find(path);
 	if (it != image_assets.end()) {
-		return it;
+		return &it.value();
```
wvpm, line 51: Why &it.value() instead of it->value() ? (as you do with asset->image)
it is a pointer to an image_asset_t, so -> should work.

Hop311: It is not pointer to an image asset, it's an iterator pointing to a key value pair, where the value is an image asset reference. `&it.value()` means taking the pointer to `it.value()` (and it has to be `it.value()`, rather than `it->second`, as only the first one gives a non-const reference, a weird part of the `ordered_map` implementation required to prevent people from editing the keys), you're thinking of `(*it).value()` which would be equivalent to `it->value()`, but `value()` is a method of the iterator, not the key-value pair you get from `*it`, so we just access it regularly and then take the pointer of the resulting image asset reference: `&(it.value())`.

wvpm: Can't you simply iterate the values only?
If you have to use a key value pair, consider calling it kvp (I'm used to seeing that in C# before we could deconstruct types). I realise now that it means iterator, it just sounds like an arbitrary name like obj for a class instance.

wvpm (approved):

extension/src/openvic-extension/singletons/AssetManager.cpp
Hop311: Nah you can't, the map stores it's entries as key value pairs and doesn't have a value-only iterator. Anywhere here I'm not iterating over the map myself, the `map.find(key)` function returns a key-value iterator. In C++ calling iterators `it` is very standard, I doubt C++ devs would know what `kvp` means, I've never seen it before myself.

PR #188, opened 2024-01-10
Reinforcements calculation

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -92,4 +92,32 @@ POPs:
...
+    formula: 5 * round(supply factor * sum of unit reinforcements * REINFORCE_SPEED * local reinforce rate * (1 + reinforce_rate) * (1 + reinforce_speed) / 5)
+    supply factor:
+      if unit received all its supplies: 1
+      else: complex graph, see https://www.desmos.com/calculator/gu1zpo01qz
+    unit reinforcements: 1000 * max_strength * unit reinforce rate
+      max_strength: as defined in /units/<unit>.txt
+      1000: (max_)strength is defined in thousands of soldiers. Reinforcements per army uses the number of soldiers.
+      unit reinforce rate:
+        if unit is mobilised: 0.5
+        else if pop is too small for a single unit: 0.5
```
Hop311, line 107: Is this if the pop is smaller than `defines.military.POP_MIN_SIZE_FOR_REGIMENT = 1000`? I believe this is also modified by:
- `defines.military.POP_MIN_SIZE_FOR_REGIMENT_PROTECTORATE_MULTIPLIER = 8`
- `defines.military.POP_MIN_SIZE_FOR_REGIMENT_COLONY_MULTIPLIER = 5`
- `defines.military.POP_MIN_SIZE_FOR_REGIMENT_NONCORE_MULTIPLIER = 3`

docs/simulation/calculations.yaml
```diff
@@ -92,4 +92,32 @@ POPs:
...
+        else if pop supports too many units: number of units it can support / number of supported units
```
Hop311, line 108: Is the number of units it can support `pop size` divided by `defines.military.POP_SIZE_PER_REGIMENT = 3000`? And does this mean that any unit requires `3000` pops, regardless of its `max_strength`?

docs/simulation/calculations.yaml
```diff
@@ -92,4 +92,32 @@ POPs:
...
+        else: 1
+    REINFORCE_SPEED: defines.military.REINFORCE_SPEED
+    local reinforce rate:
+      if location owner == army owner: 2
+      else if has military access (via treaty, occupation or allies in war): 1
+      else if is exiled: 0
+      else if location controller is hostile:
+        if location neighbours a province controller by an ally in this war: 0.5
```
Hop311, line 116:
```suggestion
        if location neighbours a province controlled by an ally in this war: 0.5
```
typo

docs/simulation/calculations.yaml
```diff
@@ -92,4 +92,32 @@ POPs:
...
+        if unit is mobilised: 0.5
+        else if pop is too small for a single unit: 0.5
+        else if pop supports too many units: number of units it can support / number of supported units
+        else: 1
+    REINFORCE_SPEED: defines.military.REINFORCE_SPEED
+    local reinforce rate:
+      if location owner == army owner: 2
+      else if has military access (via treaty, occupation or allies in war): 1
+      else if is exiled: 0
+      else if location controller is hostile:
```
Hop311, line 115: The first 2 seem to be about the location's owner, while the 4th is about its controller.
- Is a army owner owned but hostile controlled location case 1 or 4?
- Is an ally/mil access owned but hostile controlled case 2 or 4?
- What about hostile owner, but army owner or ally control?
- What about third parties, e.g. hostile owner, neutral/third party controller from a separate war?
- What does `treaty` in case 2 mean, is it subjects?

docs/simulation/calculations.yaml
```diff
@@ -92,4 +92,32 @@ POPs:
...
+        if location neighbours a province controller by an ally in this war: 0.5
+        else if location is coastal & not blockaded: 0.25
+        else: 0.1
+      else if location is uncolonised: 0
+      else: no clue
+    reinforce_rate: sum of reinforce_rate modifiers from tech
+    reinforce_speed: sum of reinforce_speed modifiers on country
+    5 * round() / 5: round to multiple of 5
```
Hop311, line 123: Shouldn't the `/ 5` be inside the `(brackets)`? Also is this rounding to the nearest integer, so `round(0.4) = 0`, `round(0.5) = 1`?

docs/simulation/calculations.yaml
wvpm: I'd rather have this in a separate section.

docs/simulation/calculations.yaml
wvpm: Yes. As above, different section imo.

docs/simulation/calculations.yaml
wvpm: Yes. Thankfully the formula has the correct variant. I simply messed up in the notes.

docs/simulation/calculations.yaml
wvpm: _Is a army owner owned but hostile controlled location case 1 or 4?_
Case 1: 200% local reinforcements. (Liberating your own provinces.)

_Is an ally/mil access owned but hostile controlled case 2 or 4?_
Case 4: 10%, 25% or 50% local reinforcements. Depending on the subcases.

_What about hostile owner, but army owner or ally control?_
Case 2: 100% local reinforcements.

_What about third parties, e.g. hostile owner, neutral/third party controller from a separate war?_
Case 2: 100% local reinforcements. Unless you're hostile to the controller, then case 4.

_What does treaty in case 2 mean, is it subjects?_
Military access via diplomacy directly or puppet/substate/puppet of substate.

Hop311: Currently case 2 says "has military access" but doesn't make clear that this is **controller** mil access, and that owner is irrelevant (mainly thinking about the second question above, which needs to be exluded from case 2). I think case 2 should specificy that:
- the location isn't owned by the army owner
- the controller isn't hostile (so is either the army owner, a war ally, treaty mil access giver, neutral third party, etc)
That way `else: no clue` shouldn't be required, as all cases are covered, every province fits into one or more of the categories:
- owned by army owner
- controlled by non-hostile
- controlled by hostile
- uncolonised

Hop311 (approved):

PR #189, opened 2024-01-11
Maximum units per soldier POP & supply factor

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -94,18 +94,31 @@ POPs:
               max(0.5: Hardcoded minimum of half the RGO income.
               2*: Hardcoded value.
 Military:
+  Maximum units per soldier POP:
+    if POP size < defines.military.POP_MIN_SIZE_FOR_REGIMENT: 0
```
wvpm, line 98: Even if location factor is between 0 and 1, this takes precedence.

docs/simulation/calculations.yaml
```diff
@@ -94,18 +94,31 @@ POPs:
...
+    else:
+      formula: ceil(POP size / (POP_SIZE_PER_REGIMENT * location factor))
+      POP_SIZE_PER_REGIMENT: defines.military.POP_SIZE_PER_REGIMENT
+      location factor:
+        if is protectorate: defines.military.POP_MIN_SIZE_FOR_REGIMENT_PROTECTORATE_MULTIPLIER
+        else if is protectorate: defines.military.POP_MIN_SIZE_FOR_REGIMENT_COLONY_MULTIPLIER
```
Hop311, line 104:
```suggestion
        else if is colony: defines.military.POP_MIN_SIZE_FOR_REGIMENT_COLONY_MULTIPLIER
```

PR #192, opened 2024-01-14
Added mod command line option

Hop311:
- Added `TYPE_PACKED_STRING_ARRAY` command line option type, which is basically just a string option that can be used multiple times without overwriting.
- Added `--mod` or `-m` option, which can be used multiple times, each time adding a mod's name (not including the `/mod/` part of its path). Mods are loaded in reverse order, so the last mod added will have the highest priority when `Dataloader` looks up files.
- Updates the SIM submodule to the `registry-template-fun` branch because that removes a `"dataloader/NodeTools.hpp"` include which was breaking the Github Action builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,10 @@ bin/*
 
 # JetBrains
 .idea
+
+# scons stuff
+.cache
+compile_commands.json
+
+# ccls
+.ccls-cache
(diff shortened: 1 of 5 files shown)


BrickPi (dismissed):
untested, code lgtm, good stuff :)

BrickPi (approved):

PR #194, opened 2024-01-15
Document <strata>_<need>_needs modifier effect

wvpm:
`<strata>_<need>_needs` scale the quantity a pop needs.
It does not affect social benefits.


Hop311 (approved):
LGTM 👍

PR #196, opened 2024-01-19
Added overlay parchment map and refactored map shader

### LINKED ISSUE #195: Search path fails with `File '' was not found`
As a result of https://github.com/OpenVicProject/OpenVic/blob/7acaf673f0465dcfa35b86bfbaf5dc0f83f2fc00/game/src/Game/GameStart.gd#L71

Better to fix it in https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/dataloader/Vic2PathSearch.cpp

Hop311:
- Changed water to use the water colormap rather than a solid blue texture. That texture is now red as a warning if it's ever accidentally used, it still needs to exist so we don't have to branch to avoid reading a terrain tile array texture for water pixels.
- Added overlay parchment/far view to the map shader, including a `parchment_mix` uniform float which controls the blending of the near (0) and far (1) map types.
- Refactored map shader to only blend terrain from the four corners of the pixel, but to only calculate a single province colour value. This creates a pixelated look along boundaries between different province colours, but that will be hidden by province/state/country borders in a future PR.
- Added `_zoom_parchment_threshold` to determine at what point to switch from the detailed terrain map type to the parchment-style map.
- Added `is_parchment_mapmode_allowed` to determine whether to allow the parchment overlay or not (when zoomed out), currently it is only false for the cosmetic terrain mapmode.
- Adjusted `MapView` zoom:
  - making the `_zoom_target_step` and `_zoom_parchment_threshold` based on the min and max.
  - making the orientation change upon zooming in kick in later and be steeper.
  - changing the Camera's `up` vector to forwards/towards the map's north, instead of straight up/out of the map, so that the Camera can face directly at the map without causing errors (`look_at` does not allow you to look along the `up` axis, as it wouldn't have a well defined rotation around that axis).
- Moved shader texture tile factor floats to `ShaderManager`, incorporating them into refactored `_set_shader_texture` and `_set_shader_asset_texture` functions.
- Updated SIM submodule to include search path fix #195, mapmode default colour adjustments, and other stuff not directly related to the Godot layer.
- Added `GUINode.` to some static function calls to stop Godot from emitting warnings.

### DIFF
--- a/extension/src/openvic-extension/singletons/GameSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/GameSingleton.cpp
@@ -62,6 +62,7 @@ void GameSingleton::_bind_methods() {
 	OV_BIND_METHOD(GameSingleton::get_mapmode_count);
 	OV_BIND_METHOD(GameSingleton::get_mapmode_identifier);
 	OV_BIND_METHOD(GameSingleton::set_mapmode, { "identifier" });
+	OV_BIND_METHOD(GameSingleton::is_parchment_mapmode_allowed);
 	OV_BIND_METHOD(GameSingleton::get_selected_province_index);
 	OV_BIND_METHOD(GameSingleton::set_selected_province, { "index" });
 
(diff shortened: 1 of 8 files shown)


Spartan322 (approved):
Tested on Linux, looks pretty good, will have to polish the zoom transition, but otherwise its great.

PR #198, opened 2024-01-19
Add Victoria 2 Folder Picker dialog

Spartan322:
Move Localization.initialize to beginning of GameStart._ready
Move compatibility mode paths search to separate _setup_compatibility_mode_paths
Move _setup_compatibility_mode_paths call to GameStart._ready
Fix `Non-argument '%s' found, skipping.` warning for Godot command line arguments

### DIFF
--- a/game/src/Game/GameStart.gd
+++ b/game/src/Game/GameStart.gd
@@ -6,13 +6,17 @@ const GameMenuScene := preload("res://src/Game/GameMenu.tscn")
 
 @export_subgroup("Nodes")
 @export var loading_screen : LoadingScreen
+@export var vic2_dir_dialog : FileDialog
 
 @export_subgroup("")
 @export var section_name : String = "general"
 @export var setting_name : String = "base_defines_path"
+
 var _settings_base_path : String = ""
+var _compatibility_path_list : PackedStringArray = []
 
 func _ready() -> void:
+	Localisation.initialize()
 	if ArgumentParser.get_argument(&"help"):
 		ArgumentParser._print_help()
 		# For some reason this doesn't get freed properly
@@ -31,7 +35,8 @@ func _ready() -> void:
 	Events.Options.load_settings_from_file()
 	sound_tab.queue_free()
 
-	loading_screen.start_loading_screen(_initialize_game)
+	await _setup_compatibility_mode_paths()
+	await loading_screen.start_loading_screen(_initialize_game)
 
 func _load_setting(file : ConfigFile) -> void:
 	if file == null: return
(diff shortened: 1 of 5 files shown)


BrickPi:
Looks good to me, could you rebase onto the reorganisation branch?

BrickPi (approved):
It's very janky on mac (fullscreen window that softlocks when minimized) but thats out of our control AFAIK, the one big thing that needs fixing for mac is that when you click cancel, the messagebox appears underneath the fullscreen window, so you can't close the program without force-quitting. If there's anything you can do about those I can re-review later, otherwise merge it, its the least of our problems on mac tbh, we need to sort out notarization / app sandboxing ASAP.

game/src/Game/GameStart.gd
```diff
@@ -67,23 +72,40 @@ func _load_compatibility_mode() -> void:
...
+			cancel_button.auto_translate = false
+			cancel_button.auto_translate = true
+			var failure_func := func() -> void:
+				get_window().mode = Window.MODE_WINDOWED
+				OS.alert(tr("ERROR_ASSET_PATH_NOT_FOUND_MESSAGE"), tr("ERROR_ASSET_PATH_NOT_FOUND"))
+				get_tree().quit()
+			vic2_dir_dialog.canceled.connect(failure_func)
+			await vic2_dir_dialog.dir_selected
+			get_tree().paused = false
+			actual_base_path = GameSingleton.search_for_game_path(vic2_dir_dialog.current_path)
```
Hop311, line 90: If the selected path doesn't work, this will default back to finding the Vic2 Steam install. For now this is fine, since if a Steam install could be found the earlier `search_for_game_path` would have got it, but in the future if we use this for a manual "switch Vic2 install directory" option we might want the ability to limit the search to only the hint path so the user sees an error message if they select an invalid folder.

game/src/Game/GameStart.gd
```diff
@@ -67,23 +72,40 @@ func _load_compatibility_mode() -> void:
...
+			if not actual_base_path:
+				failure_func.call()
+				return
 
 	if not _settings_base_path:
 		_settings_base_path = actual_base_path
 		# Save the path found in the search
 		Events.Options.save_settings_to_file()
```
Hop311, line 98: The settings base path is only set if it's empty, after that the user can't change it without editing/deleting the settings file. I don't know if command line arg paths should change it (at least after it's already set to something), as those are more likely one-off test things, but as **future additions**, I think the settings path should be changed:
- if the current path is invalid, e.g. if the Vic2 install has been moved. Currently we just accept the settings path with no validation (since we assume it has previously been validated), but a separate `validate_base_path` (which can be used by `search_for_base_path`) would be useful to checking the settings path is still valid and that base paths passed using `--base-path` are valid, so we can update the settings base path if it's no longer valid.
- we could have the settings base path listed in the options menu, editable using the interface added in this PR.

Hop311 (approved):
Great work, very clean interface 👍

game/src/Game/GameStart.gd
Spartan322: This wasn't part of this PR anyway so yeah whatever.

PR #203, opened 2024-02-04
Add issue templates

Spartan322:


valahdev:
Good initiative for reporting, we will definitely need this later on. My concern is that the user will probably make use of the wrong report for the wrong scenario, and correcting each instance with further explanation etc, will require time and effort, that can be avoided. 

Instead, we could make use of the labels, as we are currently, and merge the [Bug Report](https://github.com/OpenVicProject/OpenVic/pull/203/files#diff-7ef8d00ae79e1c2ac42d02e7069549d69e2324e6cf37f8d4373428f360091bab) with [Discrepancy Report](https://github.com/OpenVicProject/OpenVic/pull/203/files#diff-9f2009fb97104f93977b7116d0b9430b8452f953bed6cd9a5539cf5b0320d06b) into a Problem/Incident report and [Game Request](https://github.com/OpenVicProject/OpenVic/pull/203/files#diff-9b7f8f2b71e8e2f734881edb7aff8e3ad98641afdafb8741fce6d659ec4a72d0) with [Modding Request](https://github.com/OpenVicProject/OpenVic/pull/203/files#diff-da41de9e6d4446132aa5a740545720b07144dfd0b9392668011dc63519860c26) into just Request, to streamline the reporting process.

BrickPi (approved):

PR #206, opened 2024-02-11
Add workflow CI to automate update scripts

Spartan322:
These will need to be propagated once for [OpenVic-Simulation](https://github.com/OpenVicProject/OpenVic-Simulation), [OpenVic-Dataloader](https://github.com/OpenVicProject/OpenVic-Dataloader), and [lexy-vdf](https://github.com/OpenVicProject/lexy-vdf) as well, but once done so every commit to [scripts](https://github.com/OpenVicProject/scripts) master can be setup to generate a status checked auto-merge PR for each repo.

### DIFF
--- a/.github/workflows/update-scripts.yml
+++ b/.github/workflows/update-scripts.yml
@@ -1,0 +1,60 @@
+name: Update Scripts Submodule
+
+env:
+  UPDATE_BOT_NAME: 'Git bot'
+  UPDATE_BRANCH_NAME: update/scripts
+  UPDATE_BASE_BRANCH_NAME: master
+
+on:
+  # Allows you to run this workflow manually from the Actions tab or through HTTP API
+  workflow_dispatch:
+
+jobs:
+  sync:
+    name: Sync
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+    # Checkout the repository to the GitHub Actions runner
+    - name: Checkout
+      uses: actions/checkout@v4
+      with:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-scripts.yml
```diff
@@ -0,0 +1,58 @@
...
+        git submodule update --remote --recursive
+        echo "UPDATE_SCRIPTS_COMMIT=$(git rev-parse --short HEAD)" >> $GITHUB_ENV
+        cd ..
+        git checkout -b ${{ env.UPDATE_BRANCH_NAME }}
+
+    - name: Commit update
+      run: |
+        git config --global user.name ${{ env.UPDATE_BOT_NAME }}
+        git config --global user.email 'bot@noreply.github.com'
+        git remote set-url origin https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }}
```
wvpm, line 43: The only [resource](https://github.blog/2012-09-21-easier-builds-and-deployments-using-git-over-https-and-oauth/#using-oauth-with-git) I could find that does something like this says:
"Putting the token in the clone URL will result in Git writing it to the .git/config file in plain text."
And
"To avoid writing tokens to disk, don’t clone. Instead, just use the full git URL in your push/pull operations:"
The resource is from 2012 and perhaps they changed it.

wvpm (approved):

BrickPi:
Does the dependabot submodule update system cover this?

Spartan322:
Probably, I'd have to evaluate if there is anything here that has an advantage over dependabot.

PR #209, opened 2024-02-24
Added GUIScrollbar and GFXCorneredTileSupportingTexture

Hop311:
- Reworked `draw_rect_cornered` to be in its own abstract Godot class `GFXCorneredTileSupportingTexture` which both `GFXButtonStateTexture` and `GFXButtonStateHavingTexture` extend, so that button states of cornered tile textures can be drawn easily.
- Moved static GDExtension wrappers of `int_to_formatted_string` and `float_to_formatted_string` from `GameSingleton` to `GUINode`, and added `decimal_places` argument to `float_to_formatted_string`.
- Added `GUIScrollbar` which supports:
  - range limits, e.g. for min/max tariffs.
  - more and less buttons which change the value by 1, 10 (shift), 100 (control) or all the way (shift+control), and which support accelerating click and hold at a similar rate to Vic2 (60 updates per second, starting with a change of 0 per update and increasing by 1 every 10 updates).
  - Slider/grabber click and drag support, as well as track click that can turn into slider/grabber click if held and moved.
  - Hover and pressed button state textures for all sub-elements (less/more buttons, slider and track).
  - `value_changed` signal only emitted when the value actually changes, with setter functions that support blocking the signal and a `emit_value_changed` function to force the signal to be emitted (e.g. for initialising things connected to the signal).
  - There are examples in `BudgetMenu.gd`, with limits and range limits customised and `value_changed` connected to output labels.


Spartan322 (approved):
Tested on Linux

PR #210, opened 2024-03-03
Add GUIListBox + UI improvements

Hop311:
- Add `GUIListBox`, with scrollbar/scrolling and automatic child arrangment (which takes child visibility into account).
- `GUIScrollbar` tweaks:
  - length override, for `GUIListBox`'s scrollbars which need to have their lengths adjusted.
  - default start at 0 so that scrollbars start at the top rather than middle.
  - made range limits always min/max for non-range limited scrollbars.
- Connect textures' `changed` signal to `StyleBoxTexture`s' `emit_changed` function so that calling `set_icon_index` causes buttons to redraw.
- Flipped progress bar textures vertically to match Vic2, using a new `AssetManager` arg `bool flip_y`, and set their default max value to `1.0`.
- Made icons rotatable.
- Added `format_province_name` helper function.
- Fixed bug with default install directory search paths - `".."` should be `"."` to search the folder the executable is currently in.
- Reduced the number of mapmode button columns to make it take up less space (this will be replaced with the GUI generated minimap/mapmode panel soon anyway).
- Updated SIM submodule to include `UniqueKeyRegistry` invalidated identifier `std::string_view` fix (which was breaking `get_gui_position`) + technology changes needed for its nation management screen.

### DIFF
--- a/extension/src/openvic-extension/classes/GUIListBox.cpp
+++ b/extension/src/openvic-extension/classes/GUIListBox.cpp
@@ -1,0 +1,226 @@
+#include "GUIListBox.hpp"
+
+#include <godot_cpp/classes/input_event_mouse_button.hpp>
+#include <godot_cpp/variant/utility_functions.hpp>
+
+#include "openvic-extension/utility/ClassBindings.hpp"
+#include "openvic-extension/utility/UITools.hpp"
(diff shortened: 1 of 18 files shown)


BrickPi (approved):
LGTM :-)

PR #216, opened 2024-04-15
Implement Population Menu

Hop311:


Spartan322 (approved):
Tested and works on Linux

PR #217, opened 2024-04-20
Update SIM submodule

Hop311:
Update the SIM submodule 3 times, each including corresponding changes to this repo:
- SIM adds `GFX::Actor` and renames `GFX.hpp` to `GFXSprite.hpp` which requires include path changes in this repo.
- SIM changes map pixel position to province/index functions, requiring uses of those functions to be changed in this repo and the addition of `GameSingleton::get_map_dims` and `Utilities::from_godot_ivec2`.
- SIM adds support for getting the total population of a region, which is then displayed in the population menu.


Spartan322 (approved):
Tested on Linux

PR #218, opened 2024-04-20
Pre-allocate and use std::move for Godot Arrays and Dictionaries

Hop311:
- Reduces unnecessary re-allocations and copies by resizing `godot::Array`s before filling them and `std::move`-ing `godot::Array`s and `godot::Dictionary`s.
- Check return value of resize calls.


Spartan322:
Is move even necessary for a CoW collection?

Spartan322 (approved):

PR #220, opened 2024-04-20
Switch to using a single flag sheet image/texture

Hop311:
Flags are now loaded into a single large image/texture, rather than being stored as individual images.
`GFXMaskedFlagTexture` accesses flags through rects marking regions of the sheet.
3D flag models will use the texture form of the sheet, with an index passed to a shader indicating the specific flag to extract.


Spartan322 (dismissed):
Tested on Linux

PR #227, opened 2024-05-07
Models

Hop311:
- Add GDScript XAC model and XSM animation loaders.
  - This should be replaced with C++ loaders called via AssetManager (I have the basic framework of this ready, will commit to another branch once this is merged) and with more robust error/invalid input handling.
  - Supports basic materials, materials with tinted colours (used to colour soldiers according to their country, and which need improvement in shader design as the current design is capped at 32 distinct tint textures across all models) and flag material (scrolling normals + albedo from flagsheet).
  - Still needs to detect which materials have scrolling albedo textures (e.g. tank tracks, smoke), which will then be simple to add a shader for.
  - Model node structure isn't final, currently we have either a `UnitModel` custom node or a regular `Node3D` as the root with a `Skeleton3D`, meshes and animation player as children. We could instead use the `Skeleton3D` as the root and attach a script to it for models with extra features (animations, tint colours, scrolling textures), although apparently some models don't need a skeleton (such as basic buildings).
- Generates army, navy, fort and port/naval base models.
  - Currently just generates the models with animations and sets to idle, they will need a script or parent node or some other method of identication so we can connect them to the simulation, both to send info from the UI (e.g. when a unit is selected or moved) and to update the models according their state in the sim (e.g. units moving, fighting, being destroyed, and buildings being upgraded and ports/naval bases having ships move in or out).

### DIFF
--- a/extension/src/openvic-extension/register_types.cpp
+++ b/extension/src/openvic-extension/register_types.cpp
@@ -16,6 +16,7 @@
 #include "openvic-extension/singletons/GameSingleton.hpp"
 #include "openvic-extension/singletons/LoadLocalisation.hpp"
 #include "openvic-extension/singletons/MenuSingleton.hpp"
+#include "openvic-extension/singletons/ModelSingleton.hpp"
 
 using namespace godot;
 using namespace OpenVic;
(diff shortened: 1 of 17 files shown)


Spartan322 (approved):
Confirmed armies, forts, ports, and naval bases all across Africa and Europe, including ships in the relevant places. Have not checked changes to the OOB. Tested on Linux.

PR #228, opened 2024-05-07
Topbar display data + Population menu rebel icons

Hop311:
- Updates int/float to string formatting functions, in particular adding a suffix-based float formatting function as well as renaming the existing decimal place formatting function.
- Adds control code for rebel icons in the population menu, with regular rebel factions fully wired up to the simulation. Social/political movement rebels and national/flag-associated rebels still need more sim code before we can wire up their icons.
- Cleaned up population menu table sorting functions by adding some helper templates.
- Wire up all Topbar nodes at the GDScript level, using placeholder data but still with all the correct formatting functions to make the switch to actual data as easy as possible. The only unfinished bits are non-uniform label colouring (i.e. different colours in the same label), the custom £ icon (both of these issues will likely be addressed by a custom label system used throrughout the game's UI) and the line chart which doesn't yet have a node defined.


Spartan322 (approved):
Test on Linux, seems good.

PR #231, opened 2024-05-15
Fix Minimap click doing nothing

Spartan322:
Add MapView queue_province_hover_update
Fix edge scrolling continuing when mouse leaves window
Add province hover reset when mouse leaves the window
Change _edge_move_threshold to 0.025

### DIFF
--- a/game/src/Game/GameSession/MapControlPanel/Minimap.gd
+++ b/game/src/Game/GameSession/MapControlPanel/Minimap.gd
@@ -32,8 +32,8 @@ func _draw() -> void:
 # REQUIREMENTS
 # * SS-81
 # * UIFUN-127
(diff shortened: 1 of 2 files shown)


Hop311 (commented):
I think it would be better to unset the hovered province when the mouse goes out of the window, rather than leaving the last hovered province set.
Also edge scrolling continues in the direction of the last recorded `_mouse_pos_viewport` value when the mouse exits the window, better to stop edge scrolling when the mouse is out of the window (but ideally still move if it's over a menu in the window).

Hop311 (commented):
Could you increase `_edge_move_threshold` to compensate for it stopping when out of the window? Maybe to `0.025`, or whatever you feel works well.

Hop311 (approved):

PR #235, opened 2024-06-12
Update to use SIM Instance and Definition Managers

Hop311:


Spartan322 (approved):
Tested on Linux, session setup is getting to be too long.

PR #237, opened 2024-07-08
Scrolling model materials (tank tracks and smoke)

Hop311:
Tested for all vanilla models with scrolling textures. Examples can be seen by adding the following to a starting units OOB text file (and changing [the default unit animation](https://github.com/OpenVicProject/OpenVic/blob/master/game/src/Game/GameSession/ModelManager.gd#L56) to `MOVE` or `ATTACK`):
```
army = {
	name = "TANK TEST"
	location = 292
	regiment = {
		name = "Tank"
		type = tank
		home = 300
	}
}

navy = {
	name = "STEAM TEST"
	location = 2732
	ship = {
		name = "Steam Transport"
		type = steam_transport
	}
}
```

### DIFF
--- a/game/src/Game/Model/UnitModel.gd
+++ b/game/src/Game/Model/UnitModel.gd
@@ -12,21 +12,21 @@ var meshes : Array[MeshInstance3D]
 @export var primary_colour : Color:
 	set(col_in):
 		primary_colour = col_in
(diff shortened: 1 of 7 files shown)


Nemrav (approved):
Everything looks good!

PR #241, opened 2024-07-23
Cache sort order of localised population menu columns

Hop311:
Cache sort order for population menu columns sorted by localised string (pop type, culture, religion, location, and rebel faction).

### DIFF
--- a/extension/src/openvic-extension/singletons/MenuSingleton.hpp
+++ b/extension/src/openvic-extension/singletons/MenuSingleton.hpp
@@ -3,6 +3,7 @@
 #include <godot_cpp/classes/image.hpp>
 
 #include <openvic-simulation/pop/Pop.hpp>
+#include <openvic-simulation/types/IndexedMap.hpp>
 #include <openvic-simulation/types/OrderedContainers.hpp>
 
 namespace OpenVic {
@@ -63,6 +64,11 @@ namespace OpenVic {
 				SORT_LUXURY_NEEDS, SORT_REBEL_FACTION, SORT_SIZE_CHANGE, SORT_LITERACY, MAX_SORT_KEY
 			} sort_key = NONE;
 			bool sort_descending = true;
+			IndexedMap<PopType, size_t> pop_type_sort_cache;
+			IndexedMap<Culture, size_t> culture_sort_cache;
+			IndexedMap<Religion, size_t> religion_sort_cache;
+			IndexedMap<ProvinceInstance, size_t> province_sort_cache;
+			IndexedMap<RebelType, size_t> rebel_type_sort_cache;
 
 			std::vector<Pop const*> pops, filtered_pops;
 		};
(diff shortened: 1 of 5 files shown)


extension/src/openvic-extension/singletons/PopulationMenu.cpp
```diff
@@ -513,8 +510,16 @@ MenuSingleton::sort_func_t MenuSingleton::_get_population_menu_sort_func(populat
...
 
-void MenuSingleton::_population_menu_sort_pops() {
+bool MenuSingleton::_population_menu_sort_pops() {
 	if (population_menu.sort_key != population_menu_t::NONE) {
+		if (
+			!population_menu.pop_type_sort_cache.has_keys() || !population_menu.culture_sort_cache.has_keys() ||
+			!population_menu.religion_sort_cache.has_keys() || !population_menu.province_sort_cache.has_keys() ||
+			!population_menu.rebel_type_sort_cache.has_keys()
+		) {
+			ERR_FAIL_COND_V(population_menu_update_locale_sort_cache() != OK, false);
```
wvpm, line 520: What is ERR_FAIL_COND_V?
Does it conditionally return false or throw?
Is there even a path that returns false in this method?

Sorry if this is obvious to you, macros often confuse me.

extension/src/openvic-extension/singletons/PopulationMenu.cpp
```diff
@@ -525,6 +530,68 @@ void MenuSingleton::_population_menu_sort_pops() {
...
+	const auto generate_sort_cache = [this, &localised_items, &sorted_items]<HasGetIdentifier T>(
+		IndexedMap<T, size_t>& cache, std::vector<T> const& items
+	) {
+		localised_items.resize(items.size());
+		sorted_items.resize(items.size());
+
+		for (size_t idx = 0; idx < items.size(); ++idx) {
+			String identifier = std_view_to_godot_string(items[idx].get_identifier());
+			if constexpr (std::is_same_v<T, ProvinceInstance>) {
+				identifier = GUINode::format_province_name(identifier);
```
wvpm, line 555: For types that are displayed in the UI (countries, provinces, pop types, etc) maybe we should add a `get_display_name()` method that you can use here instead of type checking.

extension/src/openvic-extension/singletons/MenuSingleton.hpp
```diff
@@ -117,11 +123,12 @@ namespace OpenVic {
 		godot::Error population_menu_select_province(int32_t province_index);
 		godot::Error population_menu_toggle_expanded(int32_t toggle_index, bool emit_selected_changed = true);
 
-		void _population_menu_update_pops();
-		void _population_menu_update_filtered_pops();
+		bool _population_menu_update_pops();
```
wvpm, line 126: I don't understand what the bool indicates. Neither the function name, context or comments explain it.

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Spartan322: This a godot-cpp assert macro that does not crash, if the evaluation is true, it returns the second value.

(ERR_FAIL_COND_V)[https://github.com/godotengine/godot-cpp/blob/f2b521f55a0a2ba7080418ba898b97028da6134a/include/godot_cpp/core/error_macros.hpp#L418]

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Spartan322: This is a constexpr check, its conditionally compiled, it does not exist if T is not ProvinceInstance.

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Hop311: As Spartan said, if the first argument evaluates to false then the second argument is returned. We also use other similar macros from `godot-cpp` like `ERR_FAIL_NULL_V`, which returns its second argument if the first argument is false.

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Hop311: As Spartan said, this is a compile time check so it has no effect at runtime for types other than `ProvinceInstance`. Although we could solve this kind of scenario by giving each type a function, in this specific case it would be more trouble than it's worth as only `ProvinceInstance` has a different localisation key from its identifier (split states add a country adjective to their localised name, but that's a separate key which has to be handled at the GDExtension level (in this repo) and so couldn't be solved by adding functions to these SIM repo structs).

extension/src/openvic-extension/singletons/MenuSingleton.hpp
Hop311: It's the second case, true = no failure, false = failure. I used `bool` for functions which can't be called from GDScript, following the same pattern as we use in the SIM repo, but since these functions are all called by GDScript-accessible functions I might as well use `godot::Error` right away.

Spartan322 (approved):
Works on Linux.

PR #243, opened 2024-07-25
Music and sound effects

Nemrav:
This is a draft for the music and sound effects branch.
Does the following:
- Loads .wav sound effects by their define names, allowing them to be called from anywhere through `SoundManager.play_effect_compat("define_name",fallback_audiostream)`
- Loads and plays .mp3 music from the roots, sets up internal playlists (including titletheme for the loading screen) reads their metadata to display in the music box.

What remains to be done regarding sounds:
- The SongChances should affect what songs end up in the internal playlist once we have the simulation going.
- Adding sound effects when/where they are needed.

A special note: the dataloader was modified so that setting the roots could be done separately from the defines dataloading. This allows us to load the titletheme track first before any other files.


game/src/Game/Autoload/SoundManager.gd
```diff
@@ -9,6 +9,8 @@ var _loaded_sound : Dictionary = {}
 
 var _bus_to_stream_player : Dictionary = {}
 
+#const RATIO_FOR_LINEAR : float = 100
```
Hop311, line 12: Remove unused code?

game/src/Game/GameStart.gd
```diff
@@ -105,8 +105,35 @@ func _setup_compatibility_mode_paths() -> void:
...
+func setup_title_theme() -> void:
+	SoundSingleton.load_title_theme()
+
+	MusicConductor.setup_compat_song(SoundSingleton.title_theme)
+
+	var song_names = MusicConductor.get_all_song_names()
+
+	var title_index = song_names.find("The Coronation - Title")
+	if title_index == -1:
+		title_index = song_names.find("thecoronation_titletheme")
```
Hop311, line 129: Use `SoundSingleton.get_title_theme()` instead of hardcoding names here as well as in the C++ code?

game/src/Game/MusicConductor/MusicConductor.gd
```diff
@@ -67,24 +73,78 @@ func start_song_by_index(id: int) -> void:
...
+	
+	var title_index = song_names.find("The Coronation - Title")
+	possible_indices.remove_at(title_index)
+	
+	if len(possible_indices) < playlist_len:
+		playlist_len = possible_indices
+	
+	if len(playlist) != playlist_len:
+		playlist.resize(playlist_len)
+		playlist.fill(0)
```
Hop311, line 129: Why is this necessary, couldn't we just use `len(playlist)` everywhere and get rid of `playlist_len`? Or is it to force a max playlist length of 7?

game/src/Game/MusicConductor/MusicConductor.gd
```diff
@@ -67,24 +73,78 @@ func start_song_by_index(id: int) -> void:
...
+
+	for i in range(playlist_len):
+		var ind = randi_range(0,len(possible_indices)-1)
+		#The song we just played can be in the playlist, just not the first one
+		if i==0:
+			while(ind == last_played):
+				ind = randi_range(0,len(possible_indices)-1)
```
Hop311, line 136: This technically could cause an infinite loop, although very unlikely in practice of course.

Hop311 (commented):
Great work, main thing is to rename the addon's folder to `MusicMeta`, and then various more minor comments in the C++ and GDScript.

game/src/Game/MusicConductor/MusicConductor.gd
Nemrav: The length of 7 is probably not necessary, but I was going off this: https://github.com/joelAVS/The-Grand-Orchestra/blob/main/TGO/music/songs.txt where a comment is made about Vic2 having a bug if it couldn't build a 7 track playlist based on the songChance weights. I thought I might as well try to keep to the 7 tracks length that but avoid the bug by downsizing to meet the lowest number of actually possible indices. 

On a second look, I forget why I had the second check. Further, I think to truly 'fix' the problem vic2 had, we need something to bring the playlist length back up to 7, rather than keeping it down at the minimum.

game/project.godot
```diff
@@ -56,7 +56,7 @@ window/per_pixel_transparency/allowed=true
 
 [editor_plugins]
 
-enabled=PackedStringArray("res://addons/keychain/plugin.cfg", "res://addons/openvic-plugin/plugin.cfg")
+enabled=PackedStringArray("res://addons/MusicMeta/plugin.cfg", "res://addons/keychain/plugin.cfg", "res://addons/openvic-plugin/plugin.cfg")
```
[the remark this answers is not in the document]
Nemrav: I'm going to see if there's godot way to fix this over git, because what actually happened was I switched from one plugin called MusicMeta to another which is more active/recent called MusicMetadata (rather than the folder being renamed). That or the plugin devs made a mistake.

extension/src/openvic-extension/singletons/SoundSingleton.cpp
```diff
@@ -0,0 +1,430 @@
...
+//only used for sounds (.wav sfx)
+godot::String SoundSingleton::godot_path_to_file_name_extension(String const& path) const {
+	String name = path.replace("\\","/");
+	return name.get_slice("sound/",1); //get file name with extension
+}
+
+//Load a sound from the cache, or directly if its not in the cache
+//take in a path, extract just the file name for the cache (and defines)
+Ref<AudioStreamMP3> SoundSingleton::get_song(String const& path){
+	String name = godot_path_to_file_name(path);
```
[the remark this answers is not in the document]
Nemrav: This is to do with the SongChance and SoundEffect defines. I wanted to load the map in a way that I could access an audio file based on how it was written in the defines. Right now for example, there may be a song `song = {name = "warFolder/track1.mp3" ...}` godot_path_to_file_name lets us access this audiostream using `name` as the key in the map, rather than generating a full file path `who/knows/what/victoria II/music/warFolder/...` to access the audio stream.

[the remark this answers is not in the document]
Nemrav: Something that could be done to simplify things a bit would be to settle on using the file extension on both the music and sound folders, and combining the 2 rather similar functions. They were originally more different, but I ended up simplifying them to this.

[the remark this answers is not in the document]
Nemrav: New commit combines the functions into "to_define_file_name", with the caveat that it runs a std_to_godot_string every time is runs. Added a comment explaining why it exists.

game/src/Game/GameStart.gd
Nemrav: Resolved, but needed another method to be added to search songs by path (as opposed to by title).

Hop311 (commented):
Could you try fix the "Explode eardrums on startup" setting, which controls whether the starting music plays? I think it's broken because the music starts playing before the settings are loaded, although not sure why it doesn't pause it afterwards when it loads anyway

Hop311 (approved):
Great work 👍

PR #246, opened 2024-07-29
Search panel + text edit box UI generation

Hop311:


Spartan322 (commented):
Works on Linux but it passes move events to the map. (also reveals that we should be explicitly sending all localization strings as [utf8](https://github.com/godotengine/godot-cpp/blob/f2b521f55a0a2ba7080418ba898b97028da6134a/src/variant/char_string.cpp#L175), normal string initialization presumes [latin 1 encoding](https://en.wikipedia.org/wiki/ISO/IEC_8859-1))

Spartan322 (approved):

PR #250, opened 2024-08-01
Add movement prevention if any gui has focus

### LINKED ISSUE #248: Focused GUI elements do not consume map movement keybinds
For example if an input field has keyboard focus and the player taps the `A` key, an `a` character will appear in the input field and would move the map westward.

Spartan322:
Fixes #248 
Remove _cardinal_movement_vector function in favor of variable

Move map movement handling to _unhandled_input


Hop311 (approved):
Tested on Windows together with search-panel commit 👍

PR #252, opened 2024-08-20
Add cursor support

Nemrav:
Draft, loads cursors, animated and not animated, handles switching between cursors.
Currently contains code just for testing/demoing: right click the mouse to cycle through loaded cursors. The watch is set as an i-beam cursor so that you can try hovering over a text field (ex. in the new-game menu) to observe how the Cursor Manager handles auto-switching between animated cursors.

Loading of the cursors and generating new resolutions handled by the c++ side CursorSingleton. Managing what frame to display, and when to switch cursors handled by gdscript side CursorManager. This architecture is because we want a c++ dataloader, but we also need to listen to the scene tree to know when its safe to call a hardware cursor change.

Possible improvements:
- Doesn't support the full image spec (unhandled things like 16bit colour images that vic2 doesn't use)
- Might want to change how I handle collecting both .cur and .CUR files in the loading function (perhaps a new, "ignore-case" parameter for `.lookup_files_in_dir_recursive`?
- CursorManager creates a new `compat_Cursor` inner class whenever we switch cursors, which re-fetches the cursor stored in CursorSingleton. We can store a list of all compat_cursors to avoid this.

### DIFF
--- a/extension/doc_classes/CursorSingleton.xml
+++ b/extension/doc_classes/CursorSingleton.xml
@@ -1,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="CursorSingleton" inherits="Object" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+	</brief_description>
+	<description>
+		This singleton handles the dataloading and access to the windows .cur and .ani cursors found in the [code]/cursors[/code] folder. The functionality for setting and animating the cursors is done by the CursorManager gd script.
+	</description>
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="generate_resolution">
+			<return type="void" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="base_resolution_index" type="int" default="0" />
+			<param index="2" name="target_resolution" type="Vector2" default="Vector2(64, 64)" />
+			<description>
+				Takes the cursor image at [param base_resolution_index] then scales it to [param target_resolution] then saves it to the image as an extra resolution option.
+			</description>
+		</method>
+		<method name="get_animation_length" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns the length of the sequence for the cursor [param cursor_name]. This will be greater than or equal to the number of animation frames.
+			</description>
+		</method>
+		<method name="get_display_rates" qualifiers="const">
+			<return type="PackedFloat32Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns an array containing how long each frame should last for an animated cursor [param cursor_name]. The size of this array will be the same as the size of the sequence array.
+			</description>
+		</method>
+		<method name="get_frames" qualifiers="const">
+			<return type="ImageTexture[]" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="resolution_index" type="int" default="0" />
+			<description>
+				Returns an array of [ImageTexture] animation frames given a [param cursor_name] and a [param resolution_index].
+			</description>
+		</method>
+		<method name="get_hotspots" qualifiers="const">
+			<return type="PackedVector2Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="resolution_index" type="int" default="0" />
+			<description>
+				Returns an array of cursor click positions given a [param cursor_name] and [param resolution_index].
+			</description>
+		</method>
+		<method name="get_resolutions" qualifiers="const">
+			<return type="PackedVector2Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns an array of all the image resolutions contained in the cursor.
+			</description>
+		</method>
+		<method name="get_sequence" qualifiers="const">
+			<return type="PackedInt32Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns a list of frame indices used to produce an animation.
+			</description>
+		</method>
+		<method name="load_cursors">
+			<return type="int" enum="Error" />
+			<description>
+				Loads the cursors from the [code]/cursors[/code] folder. This function must be called before any other cursor related function.
+			</description>
+		</method>
+	</methods>
+	<members>
+		<member name="cursor_names" type="StringName[]" setter="" getter="get_cursor_names" default="[]">
+			A list of cursor file names.
+		</member>
+	</members>
+</class>
--- a/extension/src/openvic-extension/singletons/CursorSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/CursorSingleton.cpp
@@ -1,0 +1,615 @@
+#include "CursorSingleton.hpp"
+
+#include <cstdint>
+#include <vector>
+
+#include <godot_cpp/classes/dir_access.hpp>
+#include <godot_cpp/classes/input.hpp>
+#include <godot_cpp/classes/scene_tree.hpp>
+#include <godot_cpp/core/error_macros.hpp>
+#include <godot_cpp/variant/packed_float32_array.hpp>
+#include <godot_cpp/variant/packed_int32_array.hpp>
+#include <godot_cpp/variant/packed_vector2_array.hpp>
+#include <godot_cpp/variant/string.hpp>
+
+#include <godot_cpp/classes/global_constants.hpp>
+#include <godot_cpp/variant/packed_byte_array.hpp>
+#include <openvic-simulation/utility/Logger.hpp>
+#include <openvic-extension/utility/ClassBindings.hpp>
+#include <openvic-extension/utility/Utilities.hpp>
+
+using namespace godot;
+using namespace OpenVic;
+
+void CursorSingleton::_bind_methods() {
+	OV_BIND_METHOD(CursorSingleton::load_cursors);
+	OV_BIND_METHOD(CursorSingleton::get_frames, {"cursor_name","resolution_index"}, "normal", 0);
+	OV_BIND_METHOD(CursorSingleton::get_hotspots, {"cursor_name","resolution_index"}, "normal", 0);
+	OV_BIND_METHOD(CursorSingleton::get_animation_length, {"cursor_name"}, "normal");
+	OV_BIND_METHOD(CursorSingleton::get_display_rates, {"cursor_name"}, "normal");
+	OV_BIND_METHOD(CursorSingleton::get_sequence, {"cursor_name"}, "normal");
+	OV_BIND_METHOD(CursorSingleton::get_resolutions, {"cursor_name"}, "normal");
+	OV_BIND_METHOD(CursorSingleton::generate_resolution, {"cursor_name", "base_resolution_index", "target_resolution"}, "normal", 0, Vector2(64,64));
+	OV_BIND_METHOD(CursorSingleton::get_cursor_names);
+
+	ADD_PROPERTY(PropertyInfo(
+		Variant::ARRAY,
+		"cursor_names", PROPERTY_HINT_ARRAY_TYPE,
+		"StringName"),
+	"", "get_cursor_names");
+}
+
+CursorSingleton* CursorSingleton::get_singleton() {
+	return _singleton;
+}
+
+CursorSingleton::CursorSingleton() {
+	ERR_FAIL_COND(_singleton != nullptr);
+	_singleton = this;
+}
+
+CursorSingleton::~CursorSingleton() {
+	ERR_FAIL_COND(_singleton != this);
+	_singleton = nullptr;
+}
+
+TypedArray<StringName> CursorSingleton::get_cursor_names() const {
+	return cursor_names;
+}
+
+TypedArray<ImageTexture> CursorSingleton::get_frames(StringName const& name, int32_t res_index) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	std::vector<TypedArray<ImageTexture>> const& images = it->second.images;
+	ERR_FAIL_INDEX_V_MSG(res_index, images.size(), {}, vformat("Invalid image index for cursor \"%s\": %d", name, res_index));
+
+	return images[res_index];
+}
+
+PackedVector2Array CursorSingleton::get_hotspots(StringName const& name, int32_t res_index) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	std::vector<PackedVector2Array> const& hotspots = it->second.hotspots;
+	ERR_FAIL_INDEX_V_MSG(res_index, hotspots.size(), {}, vformat("Invalid hotspot index for cursor \"%s\": %d", name, res_index));
+
+	return hotspots[res_index];
+}
+
+int32_t CursorSingleton::get_animation_length(StringName const& name) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+	
+	return it->second.animation_length;
+}
+
+PackedVector2Array CursorSingleton::get_resolutions(StringName const& name) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	return it->second.resolutions;
+}
+
+PackedFloat32Array CursorSingleton::get_display_rates(StringName const& name) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	return it->second.display_rates.value_or(PackedFloat32Array());
+}
+
+PackedInt32Array CursorSingleton::get_sequence(StringName const& name) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	return it->second.sequence.value_or(PackedInt32Array());
+}
+
+void CursorSingleton::generate_resolution(StringName const& name, int32_t base_res_index, Vector2 target_res) {
+	cursor_map_t::iterator it = cursors.find(name);
+	ERR_FAIL_COND_MSG(it == cursors.end(), vformat("Cursor \"%s\" not found", name));
+	cursor_asset_t& cursor = it.value();
+
+	ERR_FAIL_INDEX_MSG(
+		base_res_index, cursor.images.size(), vformat("Invalid image index for cursor \"%s\": %d", name, base_res_index)
+	);
+
+	TypedArray<ImageTexture> const& images = cursor.images[base_res_index];
+	PackedVector2Array const& hotspots = cursor.hotspots[base_res_index];
+	const Vector2 resolution = cursor.resolutions[base_res_index];
+	TypedArray<ImageTexture> new_frameset;
+	PackedVector2Array new_hotspots;
+
+	for (size_t index = 0; index < images.size(); index++) {
+		Ref<ImageTexture> const& texture = images[index];
+		Ref<Image> image;
+		image.instantiate();
+		image->copy_from(texture->get_image());
+		image->resize(target_res.x, target_res.y, Image::INTERPOLATE_BILINEAR);
+		new_frameset.push_back(ImageTexture::create_from_image(image));
+		new_hotspots.push_back((hotspots[index] * target_res / resolution).floor());
+	}
+
+	cursor.images.push_back(new_frameset);
+	cursor.hotspots.push_back(new_hotspots);
+	cursor.resolutions.push_back(target_res);
+}
+
+static constexpr std::string_view cursor_directory = "gfx/cursors";
+
+static String _to_define_file_name(String const& path) {
+	static const String backslash = "\\";
+	static const String forwardslash = "/";
+	static const String cursor_directory_forwardslash = Utilities::std_to_godot_string(cursor_directory) + forwardslash;
+	static const String dot = ".";
+	return path.replace(backslash, forwardslash).get_slice(cursor_directory_forwardslash, 1).get_slice(dot, 0);
+}
+
+Error CursorSingleton::load_cursors() {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V_MSG(game_singleton, FAILED, "Error retrieving GameSingleton");
+
+	//there is also a png file in the folder we don't want to bother loading
+	//so don't just load everything in the directory
+
+	//We need to load both ".cur" and ".CUR" files
+	Dataloader::path_vector_t cursor_files = game_singleton->get_dataloader()
+		.lookup_files_in_dir_recursive(cursor_directory, ".cur");
+
+	Dataloader::path_vector_t CURsor_files = game_singleton->get_dataloader()
+		.lookup_files_in_dir_recursive(cursor_directory, ".CUR");
+	cursor_files.insert(std::end(cursor_files),std::begin(CURsor_files),std::end(CURsor_files));
+
+	Dataloader::path_vector_t animated_cursor_files = game_singleton->get_dataloader()
+		.lookup_files_in_dir_recursive(cursor_directory, ".ani");
+	
+	if (cursor_files.empty() && animated_cursor_files.empty()){
+		Logger::error("failed to load cursors: no files in cursors directory");
+		return FAILED;
+	}
+
+	Error ret = OK;
+
+	for(fs::path const& file_name : cursor_files) {
+		String file = Utilities::std_to_godot_string(file_name.string());
+		StringName name = _to_define_file_name(file);
+
+		if (!_load_cursor_cur(name,file)){
+			Logger::error("failed to load normal cursor at path ", file_name);
+			ret = FAILED;
+		}
+	}
+
+	for(fs::path const& file_name : animated_cursor_files) {
+		String file = Utilities::std_to_godot_string(file_name.string());
+		StringName name = _to_define_file_name(file);
+
+		if (!_load_cursor_ani(name,file)){
+			Logger::error("failed to load animated cursor at path ", file_name);
+			ret = FAILED;
+		}
+	}
+
+	return ret;
+}
+
+static constexpr int32_t _reverser_lookup[] {
+	0x0, 0x8, 0x4, 0xc, 0x2, 0xa, 0x6, 0xe,
+	0x1, 0x9, 0x5, 0xd, 0x3, 0xb, 0x7, 0xf
+};
+static constexpr int32_t _reverse_bits(int32_t byte, int32_t bits_per_pixel=8) {
+	int32_t a = _reverser_lookup[(byte & 0b1111)] << 4;
+	int32_t b = _reverser_lookup[byte >> 4];
+	int32_t c = b | a;
+	return c >> (8-bits_per_pixel);
+}
+
+static constexpr int32_t _rotate_right(int32_t byte, int32_t size=8) {
+	return ((byte & 0b1) << (size-1)) | (byte >> 1);
+}
+
+static int32_t _load_int_256(Ref<FileAccess> const& file) {
+	int32_t value = file->get_8();
+	if (value == 0) value = 256;
+	return value;
+}
+
+static constexpr int32_t _get_row_start(int32_t x_coord, int32_t y_coord, int32_t bits_per_pixel) {
+	x_coord *= bits_per_pixel;
+	int32_t row_count = (x_coord + 31) >> 5;
+	return row_count * y_coord * 4; // 4 bytes per row * rows down
+}
+
+static constexpr int32_t _select_bits(uint8_t const* data, int32_t row_start, int32_t first_bit, int32_t bit_count) {
+	int32_t byte_index = first_bit >> 3;
+	int32_t bit_in_byte_index = first_bit & 0b111;
+	if (bit_in_byte_index + bit_count > 8) {
+		Logger::error("Attempted to select bits outside of a byte.");
+		return 0;
+	}
+	int32_t byte = _reverse_bits(*(data+row_start+byte_index));
+	int32_t selected = (byte >> bit_in_byte_index) & ((1 << bit_count) - 1);
+
+	//TODO: questionable hack, nothing in the spec suggests we should need to do this
+	if (bit_count > 1 && selected != 0){
+		return _rotate_right(selected,4);
+	}
+	return selected;
+}
+
+static constexpr bool _read_AND_mask(
+	uint8_t const* data, int32_t pixel_x, int32_t pixel_y, int32_t x_dimension, int32_t offset) {
+	
+	int32_t row_start = _get_row_start(x_dimension, pixel_y, 1);
+	int32_t and_bit = _select_bits(data, row_start + offset,pixel_x, 1);
+	return !and_bit;
+}
+
+static void _pixel_palette_lookup(
+		PackedByteArray const& data, PackedByteArray& pixel_data, uint32_t i,
+		PackedByteArray const& palette, int32_t coord_x, int32_t coord_y, int32_t x_dimension,
+		int32_t offset, bool transparent, int32_t bits_per_pixel) {
+	
+	int32_t row_start = _get_row_start(x_dimension, coord_y, bits_per_pixel);
+	int32_t pixel_bits = _select_bits(data.ptr(), row_start + offset, coord_x*bits_per_pixel, bits_per_pixel);
+	
+	if ((pixel_bits+1)*4 > palette.size()){
+		Logger::error("attempted to select invalid colour palette entry, ", pixel_bits);
+		return;
+	}
+	
+	//pixel bits serves as an index into the colour palette. We need to multiply the index by the number of bytes per colour (4)
+	pixel_data[(i*4) + 0] = palette[pixel_bits*4 + 0];
+	pixel_data[(i*4) + 1] = palette[pixel_bits*4 + 1];
+	pixel_data[(i*4) + 2] = palette[pixel_bits*4 + 2];
+	pixel_data[(i*4) + 3] = 0xFF * transparent; //a
+}
+
+/*
+24bit pixel support here is questionable:
+the spec (per daubnet) says we should pad bytes to end things on 32bit boundaries
+but the singular example of a 24bit cursor found on the internet does things like this.
+So emit a warning when trying to load one of these
+*/
+static void _read_24bit_pixel(
+	PackedByteArray const& image_data, PackedByteArray& pixel_data,
+	int32_t i, int32_t offset, bool opaque) {
+
+	if((i+1)*3 > image_data.size()){
+		Logger::error("Pixel ", i, "tried to read from a pixel data array of max size ", pixel_data.size());
+		return;
+	}
+
+	pixel_data[(i*4) + 0] = image_data[offset + (i*3) + 2]; //r
+	pixel_data[(i*4) + 1] = image_data[offset + (i*3) + 1]; //g
+	pixel_data[(i*4) + 2] = image_data[offset + (i*3) + 0]; //b
+	pixel_data[(i*4) + 3] = 0xFF * opaque; //a
+}
+
+static void _read_32bit_pixel(
+	PackedByteArray const& image_data, PackedByteArray& pixel_data,
+	int32_t i, int32_t offset, bool opaque) {
+
+	if((i+1)*4 > image_data.size()){
+		Logger::error("Pixel ", i, "tried to read from a pixel data array of max size ", pixel_data.size());
+		return;
+	}
+
+	pixel_data[(i*4) + 0] = image_data[offset + (i*4) + 2]; //r
+	pixel_data[(i*4) + 1] = image_data[offset + (i*4) + 1]; //g
+	pixel_data[(i*4) + 2] = image_data[offset + (i*4) + 0]; //b
+	pixel_data[(i*4) + 3] = image_data[offset + (i*4) + 3] * opaque; //a
+}
+
+//used to load a .cur file from a file (could be the a whole .cur file, or a .cur within a .ani file)
+static CursorSingleton::image_hotspot_pair_asset_t _load_pair(Ref<FileAccess> const& file) {
+	CursorSingleton::image_hotspot_pair_asset_t pairs = {};
+
+	//.cur's within .anis won't start of the beginning of the file, so save where they start
+	int32_t base_offset = file->get_position();
+
+	//.cur header
+	int32_t reserved = file->get_16();
+	int32_t type = file->get_16(); //1=ico, 2=cur
+	int32_t images_count = file->get_16();
+	
+	//all the images
+	for(int32_t i=0; i<images_count; i++){
+		Vector2i dimensions = Vector2i(_load_int_256(file),_load_int_256(file)); //TODO
+		int32_t palette = file->get_8();
+		file->get_8(); //int32_t img_reserved
+		
+		Vector2i hotspot = Vector2i();
+		hotspot.x = file->get_16();
+		hotspot.y = file->get_16();
+
+		int32_t data_size = std::min(static_cast<uint64_t>(file->get_32()), file->get_length() - file->get_position());
+		int32_t data_offset = file->get_32();
+
+		//This image header information is sequential in the data, but the images aren't necessarily
+		// so save the current position, get the image data and return so we're ready for the next image header
+		int32_t end_of_image_header = file->get_position();
+
+		file->seek(data_offset+base_offset);
+		PackedByteArray const& image_data = file->get_buffer(data_size);
+		file->seek(end_of_image_header);
+
+		Ref<Image> image = Ref<Image>();
+		image.instantiate();
+	
+		//PNGs are stored in their entirety, so use Godot's internal loader
+		if (image_data.slice(1,4).get_string_from_ascii() == "PNG") {
+			image->load_png_from_buffer(image_data);
+		}
+		else { //BMP based cursor, have to load this manually
+			//int32_t dib_header_size = image_data.decode_u32(0);
+
+			//this is the combined sized of the picture and the transparency bitmask
+			// (ex. 32x32 dimension image becomes 32x64 here)
+			//Vector2i combined_dimensions = Vector2i(image_data.decode_u32(4),image_data.decode_u32(8));
+			//int32_t colour_planes = image_data.decode_u16(12);
+			int32_t bits_per_pixel = image_data.decode_u16(14);
+			if (bits_per_pixel <= 8 || bits_per_pixel == 24){
+				Logger::warning("Attempting to import ", bits_per_pixel, "bit cursor, this isn't guaranteed to work");
+			}
+			else if (bits_per_pixel != 32){
+				Logger::error("Invalid or Unsupported bits per pixel while loading cursor image, bpp: ", bits_per_pixel, "loading blank image instead");
+			}
+
+			int32_t size = image_data.decode_u32(20);
+			Vector2i resolution = Vector2i(image_data.decode_s32(24),image_data.decode_s32(28));
+			int32_t palette_size = image_data.decode_u32(32);
+
+			if (palette_size == 0 && bits_per_pixel <= 8){
+				palette_size =  1 << bits_per_pixel;
+			}
+			//int32_t important_colours = image_data.decode_u32(36);
+
+			//for BMPs with 8 bits per pixel or less, the pixel data is actually a lookup to this table here
+			PackedByteArray const& palette = image_data.slice(40,40+(4*palette_size));
+
+			// this is where the image data starts
+			int32_t offset = 40 + palette_size*4;
+
+			//where the transparency AND mask starts
+			int32_t mask_offset = offset + _get_row_start(dimensions.x,dimensions.y,bits_per_pixel);
+
+			PackedByteArray pixel_data = PackedByteArray();
+			pixel_data.resize(dimensions.x*dimensions.y*4);
+			pixel_data.fill(255);
+
+			int32_t i=0;
+			for(int32_t row=0; row < dimensions.y; row++) {
+				for(int32_t col=0; col < dimensions.x; col++) {
+					Vector2i coord = Vector2i(col,row);
+					bool transparent = _read_AND_mask(
+						image_data.ptr(),coord.x,coord.y,dimensions.x,mask_offset
+					);
+					if (bits_per_pixel <= 8){
+						//mostly legacy files, these ones all use a lookup into the colour palette
+						_pixel_palette_lookup(
+							image_data, pixel_data, i, palette, coord.x, coord.y, dimensions.x, offset, transparent, bits_per_pixel
+						);
+					}/*
+					else if (bits_per_pixel == 16) { //TODO
+						//Unsupported, error
+					}*/
+					else if (bits_per_pixel == 24) {
+						//Support Questionable, based on 1 example on the internet as opposed to the actual spec
+						_read_24bit_pixel(
+							image_data, pixel_data, i, offset, transparent
+						);
+					}
+					else if (bits_per_pixel == 32) {
+						//What vic actually uses
+						_read_32bit_pixel(
+							image_data, pixel_data, i, offset, transparent
+						);
+					}
+					i++;
+				}
+			}
+			
+			image = image->create_from_data(dimensions.x,dimensions.y,false, Image::FORMAT_RGBA8,pixel_data);
+			//bmp images are stored bottom to top
+			image->flip_y();
+		}
+		Ref<ImageTexture> image_texture = Ref<ImageTexture>();
+		image_texture.instantiate();
+
+		image_texture = image_texture->create_from_image(image);
+		
+		if (image_texture.is_null()){
+			Logger::error("Image Texture ",Utilities::godot_to_std_string(file->get_path())," was null!");
+		}
+
+		pairs.hotspots.push_back(hotspot);
+		pairs.images.push_back(image_texture);
+		
+	}
+	return pairs;
+
+}
+
+bool CursorSingleton::_load_cursor_ani(StringName const& name, String const& path) {
+	const Ref<FileAccess> file = FileAccess::open(path, FileAccess::ModeFlags::READ);
+
+	const Error err = FileAccess::get_open_error();
+	ERR_FAIL_COND_V_MSG(
+		err != OK || file.is_null(), false, vformat("Failed to open ani file: \"%s\"", path)
+	);
+
+	//read the RIFF container
+	Utilities::read_riff_str(file); //riff_id
+	const uint64_t riff_size = std::min(static_cast<uint64_t>(file->get_32()), file->get_length());;
+	Utilities::read_riff_str(file); //form_type
+
+	//important variables
+	std::vector<TypedArray<ImageTexture>> frames_by_resolution;
+	std::vector<PackedVector2Array> hotspots_by_resolution;
+
+	PackedVector2Array resolutions;
+	PackedFloat32Array display_rates;
+	PackedInt32Array sequence;
+
+	//ani header variables
+	int32_t num_frames = 1;
+	int32_t num_steps = 1;
+	Vector2i dimensions = Vector2i(1,1);
+	int32_t bit_count = 1;
+	int32_t num_planes = 1; //???
+	int32_t display_rate = 1; //how long each frame should last
+	int32_t flags = 0;
+	bool icon_flag = false;
+	bool sequence_flag = false;
+
+
+	while(file->get_position() < riff_size){
+		String id = Utilities::read_riff_str(file);
+		int32_t size = file->get_32();
+		if (id == "LIST"){
+			String list_type = Utilities::read_riff_str(file);
+		}
+		else if (id == "anih"){
+			//hack for some files, there's likely a better way
+			if (size == 36){ 
+				int32_t headerSize = file->get_32();
+			}
+			num_frames = file->get_32();
+			num_steps = file->get_32();
+			dimensions = Vector2i(file->get_32(),file->get_32());
+			bit_count = file->get_32();
+			num_planes = file->get_32();
+			display_rate = file->get_32();
+			flags = file->get_32();
+			icon_flag = flags & 0x1;
+			sequence_flag = flags & 0x2;
+		}
+		else if (id == "icon"){
+
+			int32_t file_access_offset = file->get_position();
+
+			image_hotspot_pair_asset_t pair = _load_pair(file);
+			//basically pushback an array
+
+			//only store the resolutions from one frame
+			if (resolutions.is_empty()){
+				for(int32_t i=0;i<pair.images.size();i++){
+					
+					PackedVector2Array hotspots;
+					TypedArray<ImageTexture> images;
+					images.push_back(pair.images[i]);
+					hotspots.push_back(pair.hotspots[i]);
+					frames_by_resolution.push_back(images);
+					hotspots_by_resolution.push_back(hotspots);
+
+					resolutions.push_back(Vector2(pair.images[i]->get_width(),pair.images[i]->get_height()));
+				}
+
+			}
+			else {
+				if (pair.images.size() != frames_by_resolution.size()){
+					Logger::error(
+						"Malformatted .ani cursor file ",
+						Utilities::godot_to_std_string(name),
+						" had inconsistent number of images per cursor"
+					);
+				}
+				for(int32_t i=0; i<pair.images.size(); i++){
+					frames_by_resolution[i].push_back(pair.images[i]);
+					hotspots_by_resolution[i].push_back(pair.hotspots[i]);
+				}
+			}
+
+			//cursor could have been anywhere in the file, come back to a known position
+			file->seek(file_access_offset + size);
+
+		}
+		else if (id == "seq "){
+			for(int32_t i=0; i<num_steps; i++){
+				sequence.push_back(file->get_32());
+			}
+		}
+		else if (id == "rate"){
+			for(int32_t i=0;i<num_steps;i++){
+				display_rates.push_back(file->get_32()/60.0);
+			}
+		}
+		else {
+			//Various junk (JUNK, metadata we don't care about, ...)
+			file->get_buffer(size);
+		}
+		//align to even bytes
+		if ((file->get_position() & 1) != 0){
+			file->get_8();
+		}
+	}
+
+	//not all ani files have the sequence and rate chunks, if not, fill out these properties
+	//manually
+	if (sequence.is_empty()){
+		for(int32_t i=0; i<num_steps;i++){
+			sequence.push_back(i);
+		}
+	}
+	if (display_rates.is_empty()){
+		for(int32_t i=0; i<num_steps;i++){
+			display_rates.push_back(display_rate/60.0);
+		}
+	}
+
+	cursors.emplace(
+		name,
+		cursor_asset_t {
+			std::move(hotspots_by_resolution),
+			std::move(frames_by_resolution),
+			resolutions,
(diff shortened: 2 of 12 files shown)


extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,99 @@
...
+		//.cur files use all but the last 2 properties, rest are for .ani
+		//use vector2i instead of ref<vector2i> because of "unreference is not a member of godot::vector2i"
+		//images is an array of ImageTexture, just godot doesn't like typed arrays with image texture for some reason
+		struct cursor_asset_t {
+			std::vector<godot::TypedArray<godot::Vector2i>> hotspots;
+			std::vector<godot::Array> images;
+			godot::TypedArray<godot::Vector2i> resolutions;
+			int animationLength; //1 for static cursors
+			std::optional<godot::TypedArray<float>> displayRates;
+			std::optional<godot::TypedArray<int>> sequence;
```
Hop311, line 47: Godot has special `Packed***Array` types that are faster and more space-efficient than regular arrays full of the same value type:
- `PackedByteArray` - array of `uint8_t`s
- `PackedColorArray` - array of `Color`s (so basically `Vector4`s)
- `PackedFloat32Array` - array of `float`s
- `PackedFloat64Array` - array of `double`s
- `PackedInt32Array` - array of `int32_t`s
- `PackedInt64Array` - array of `int64_t`
- `PackedStringArray` - array of `godot::String`s
- `PackedVector2Array` - array of `Vector2`s (2 floats)
- `PackedVector3Array` - array of `Vector3`s (3 floats)
- `PackedVector4Array` - array of `Vector4`s (4 floats)

So here it would be best to use `godot::PackedFloat32Array` and `godot::PackedInt32Array`. Unfortunately there isn't a `PackedVector2iArray` we could use for `resolutions`, so that will have to stay as a regular array (unless you're able to switch to using float-based `Vector2`s? Tbh probably not worth it anyway).

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+			else if(bitsPerPixel != 32){
+				Logger::error("Invalid or Unsupported bits per pixel while loading cursor image, bpp: ", bitsPerPixel, "loading blank, transparent image instead");
+			}
+
+			int size = imageData.decode_u32(20);
+			Vector2i resolution = Vector2i(imageData.decode_s32(24),imageData.decode_s32(28));
+			int paletteSize = imageData.decode_u32(32);
+
+			if(paletteSize == 0 && bitsPerPixel <= 8){
+				paletteSize =  static_cast<int>(pow(2, bitsPerPixel));
```
Hop311, line 409: Rather than `pow(2, x)` and float/int casting, you can just do `1 << x`

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+	pixel.append(b);
+	pixel.append(a);
+
+	return pixel;
+}
+
+int CursorSingleton::_get_row_start(Vector2i dimensions, int y_coord, int bitsPerPixel){
+	int rowCount = dimensions.x * bitsPerPixel / 32;
+	bool hasExtraRow = ((dimensions.x * bitsPerPixel) % 32) != 0;
+	rowCount += 1*hasExtraRow;
```
Hop311, line 545: Remove the `1 * ...` ?

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+	//ani header variables
+	int numFrames;
+	int numSteps;
+	Vector2i dimensions;
+	int bitCount;
+	int numPlanes; //???
+	int displayRate; //how long each frame should last
+	int flags;
+	bool iconFlag;
+	bool sequenceFlag;
```
Hop311, line 209: With MSVC primitive type variables aren't always default initialised, so these could contain random data. Always safer to explicitly initialise them, especially if there's any chance they'll be read later without having been changed.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+//, std::string_view const& base_folder
+String CursorSingleton::to_define_file_name(String const& path) const {
+	String name = path.replace("\\","/");
+	return name.get_slice("gfx/cursors/",1).get_slice(".",0);
+}
+
+bool CursorSingleton::load_cursors() {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V_MSG(game_singleton, false, vformat("Error retrieving GameSingleton"));
```
Hop311, line 131: `vformat` isn't needed if there's nothing to be substituted in (using the same `%` system as GDScript)

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+TypedArray<float> CursorSingleton::get_displayRates(String const& name){
+	if(cursors[name].displayRates.has_value()){
+		return cursors[name].displayRates.value();
+	}
+	return TypedArray<float>();
+}
+
+TypedArray<int> CursorSingleton::get_sequence(String const& name){
+	if(cursors[name].displayRates.has_value()){
```
Hop311, line 84:
```suggestion
	if(cursors[name].sequence.has_value()){
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+Array CursorSingleton::get_frames(String const& name, int res_index){
+	if(res_index < cursors[name].images.size()){
+		return cursors[name].images[res_index];
+	}
+	return Array();
+}
+
+TypedArray<Vector2i> CursorSingleton::get_hotspots(String const& name, int res_index){
+	if(res_index < cursors[name].images.size()){
```
Hop311, line 62:
```suggestion
	if(res_index < cursors[name].hotspots.size()){
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,586 @@
...
+
+#include <openvic-extension/utility/ClassBindings.hpp>
+#include <openvic-extension/utility/Utilities.hpp>
+
+using OpenVic::Utilities::godot_to_std_string;
+using OpenVic::Utilities::std_to_godot_string;
+
+using namespace godot;
+using namespace OpenVic;
+using namespace OpenVic::NodeTools;
```
Hop311, line 19: Is this needed in this file?

extension/src/openvic-extension/singletons/SoundSingleton.cpp
```diff
@@ -1,7 +1,7 @@
 #include "SoundSingleton.hpp"
 
 #include <string_view>
-#include <vector>
+//#include <vector>
```
Hop311, line 4: Was this added to the PR by accident?

game/src/Game/GameStart.gd
```diff
@@ -107,9 +107,16 @@ func _setup_compatibility_mode_paths() -> void:
...
+	#CursorSingleton.load_cursors()
+	CursorManager.set_prefered_res(Vector2i(48,48))
+	CursorManager.load_cursors()
+	CursorManager.set_compat_cursor("normal",Input.CURSOR_ARROW)
+	#TODO: Change Busy to be shape Input.CURSOR_BUSY when done testing
+	CursorManager.set_compat_cursor("busy",Input.CURSOR_IBEAM)
+	
 	setup_title_theme()
-
+
```
Hop311, line 119: Stray tab got into the PR

game/src/Game/GameStart.gd
```diff
@@ -107,9 +107,16 @@ func _setup_compatibility_mode_paths() -> void:
 func _load_compatibility_mode() -> void:
 	if GameSingleton.set_compatibility_mode_roots(_compatibility_path_list) != OK:
 		push_error("Errors setting game roots!")
-
+	
+	#CursorSingleton.load_cursors()
+	CursorManager.set_prefered_res(Vector2i(48,48))
+	CursorManager.load_cursors()
+	CursorManager.set_compat_cursor("normal",Input.CURSOR_ARROW)
```
Hop311, line 114: The cursor names should probably be StringNames, both as literals (`&"normal"`) and the variable/argument types that eventually are used to call `CursorSingleton` functions which accept a `StringName`

Hop311 (commented):
Some random comments from looking through the code (not form proper testing yet)

extension/src/openvic-extension/singletons/CursorSingleton.hpp
Nemrav: Made the change to PackedFloat and Ints. You probably could use Vector2s instead of Vector2is, so I'll leave this unresolved in case you really do want that.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
Nemrav: We do use godot for various datatypes. Namespace OpenVic is used for GameSingleton access only to my knowledge.
Removed NodeTools because it is completely unused.

game/src/Game/MusicConductor/MusicConductor.gd
Nemrav, line 1: This was a fix for a situation where no compat songs could be loaded. I should move this to a separate pr.

extension/src/openvic-extension/singletons/SoundSingleton.cpp
Nemrav: removed

game/src/Game/GameStart.gd
Nemrav: Made this change.

Nemrav:
Main concern for the review is the initial loading screen: 
- is it fine that the loading screen initially uses the normal arrow cursor until the player's mouse moves to cause an update to the "busy" (loading) cursor. I think this might just be down to godot.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,585 @@
...
+		Vector2i base_hotspot = cursor->hotspots[base_res_index][i];
+		Vector2i scale_ratio = target_res / static_cast<Vector2i>(cursor->resolutions[base_res_index]);
+		new_hotspots.push_back(base_hotspot*scale_ratio);
+
+	}
+
+	cursor->images.push_back(new_frameset);
+	cursor->hotspots.push_back(new_hotspots);
+	cursor->resolutions.push_back(target_res);
+}
```
Hop311, line 120:
```suggestion
void CursorSingleton::generate_resolution(StringName const& name, int32_t base_res_index, Vector2i target_res) {
	cursor_map_t::iterator it = cursors.find(name);
	ERR_FAIL_COND_MSG(it == cursors.end(), vformat("Cursor \"%s\" not found", name));

	cursor_asset_t& cursor = it.value();

	ERR_FAIL_INDEX_MSG(
		base_res_index, cursor.images.size(), vformat("Invalid image index for cursor \"%s\": %d", name, base_res_index)
	);

	TypedArray<ImageTexture> const& images = cursor.images[base_res_index];
	TypedArray<Vector2i> const& hotspots = cursor.hotspots[base_res_index];
	const Vector2i resolution = cursor.resolutions[base_res_index];

	TypedArray<ImageTexture> new_frameset;
	TypedArray<Vector2i> new_hotspots;

	for (size_t index = 0; index < images.size(); index++) {
		Ref<ImageTexture> const& texture = images[index];

		Ref<Image> image;
		image.instantiate();
		image->copy_from(texture->get_image());
		image->resize(target_res.x, target_res.y, Image::INTERPOLATE_BILINEAR);

		new_frameset.push_back(ImageTexture::create_from_image(image));

		new_hotspots.push_back(static_cast<Vector2i>(hotspots[index]) * target_res / resolution);
	}

	cursor.images.push_back(new_frameset);
	cursor.hotspots.push_back(new_hotspots);
	cursor.resolutions.push_back(target_res);
}
```
- `StringName`, `int32_t` and `TypedArray<ImageTexture>` as mentioned in header comments.
- `find` for looking up in `cursors` to avoid crash on missing entry. The local `cursor` variable can be a reference rather than a pointer as it'll never be null. We need to use `it.value()` here rather than `it->second` as we need a non-const reference (it's a weird quirk of the map containers we use, when accessed as a `(key, value)` pair they're always const to protect the key from being modified, but we can still get the non-const value via the `it.value()` helper function).
- Error check on `base_res_index` (we only need to check against one of the vectors, assuming images, hotspots and resolutions are always grown in tandem).
- Local `images`, `hotspots` and `resolution` variables to avoid unnecessary extra lookups in each iteration of the for loop.
- Better to use `size_t index` as it matches the type of `images.size()` and it's generally better to avoid single-letter variable names, both for clarity and to avoid typos.
- There's no need to call `image->create`, `copy_from` overrides any existing data within the image (we still need `instantiate` to allocate memory for the `Image`'s internal variables before they can then be set by `copy_from`).
- We can use the static helper function `ImageTexture::create_from_image` rather than manually instantiating the `ImageTexture` and setting its `Image`.
- Better to do `static_cast<Vector2i>(hotspots[index]) * target_res / resolution` rather than `static_cast<Vector2i>(hotspots[index]) * (target_res / resolution), it reduces the integer division error, e.g. `6 * 4 / 3 = 24 / 3 = 8` while `6 * (4 / 3) = 6 * 1 = 6`. The `static_cast<Vector2i>` wouldn't be necessary if we were using a `Packed...Array`, maybe we could use `PackedVector2Array` but that might require some extra code to ensure the new hotspot is made up of whole numbers, e.g. by using `Vector2`'s `floor` function.

Hop311 (commented):
Great work 🎉
Could you address the comments I've made + rebase onto the current master branch? I'll continue reviewing tomorrow after work.
Also I noticed that in the province search panel, the cursor goes to the normal `I` shape when hovering over the text edit box, while in Vic2 it stays as the pointing gauntlet.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,585 @@
...
+	_singleton = nullptr;
+}
+
+
+Array CursorSingleton::get_frames(String const& name, int res_index){
+	if(res_index < cursors[name].images.size()){
+		return cursors[name].images[res_index];
+	}
+	return Array();
+}
```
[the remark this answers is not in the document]
Nemrav: value_or({}); gives an error, but value_or(PackedInt32Array()); seems to work.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
Nemrav: Since hotspots was changed to a PackedVector2Array, I got rid of the cast, and added .floor() as suggested.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,602 @@
...
+	cursor_asset_t cursor = {
+		hotspots_by_resolution,
+		frames_by_resolution,
+		resolutions,
+		static_cast<int>(sequence.size())
+	};
+
+	cursor.displayRates = displayRates;
+	cursor.sequence = sequence;
+	cursors.emplace(std::move(name),cursor);
```
Hop311, line 316:
```suggestion
	cursors.emplace(
		name,
		cursor_asset_t {
			std::move(hotspots_by_resolution),
			std::move(frames_by_resolution),
			resolutions,
			static_cast<int32_t>(sequence.size()),
			displayRates,
			sequence
		}
	);
```
- We can `std::move` the `hotspots_by_resolution` and `frames_by_resolution` vectors into the `cursor_asset_t` rather than copying them.
- We can include `displayRates` and `sequence` in the intialiser list rather than setting then manually after construction.
- We could `std::move` `cursor` into `cursors` to avoid copying it (particularly its vector member variables), although it would be simplest to just construct it in place.
- We can remove the `std::move` on `name` as it has no effect given `name` is a `const&` variable (forces a copy as there's no mutable instance to steal allocated memory from) and anyway with `StringName`s there's nothing saved by moving.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,602 @@
...
+	TypedArray<Vector2> new_hotspots;
+
+	for (size_t index = 0; index < images.size(); index++) {
+		Ref<ImageTexture> const& texture = images[index];
+		Ref<Image> image;
+		image.instantiate();
+		image->copy_from(texture->get_image());
+		image->resize(target_res.x, target_res.y, Image::INTERPOLATE_BILINEAR);
+		new_frameset.push_back(ImageTexture::create_from_image(image));
+		new_hotspots.push_back((hotspots[index].floor()) * target_res / resolution);
```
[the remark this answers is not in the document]
Nemrav: Changed, but leaving as unresolved so we remember to check.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,622 @@
...
+	if (value == 0) value = 256;
+	return value;
+}
+
+static int32_t _get_row_start(Vector2i dimensions, int32_t y_coord, int32_t bitsPerPixel) {
+	int32_t rowCount = dimensions.x * bitsPerPixel / 32;
+	bool hasExtraRow = ((dimensions.x * bitsPerPixel) % 32) != 0;
+	rowCount += hasExtraRow; //TODO 1*
+	return rowCount*y_coord*4; //4 bytes per row * rows down
+}
```
Hop311, line 220:
```suggestion
static constexpr int32_t _get_row_start(int32_t x_coord, int32_t y_coord, int32_t bitsPerPixel) {
	x_coord *= bitsPerPixel;
	int32_t rowCount = (x_coord + 31) >> 5;
	return rowCount * y_coord * 4; // 4 bytes per row * rows down
}
```
- On further inspection, this function can also be `constexpr` if we just replace the `Vector2i` with an `int32_t` (since we only use its x coordinate anyway) and then just use `dimensions.x` as its first argument later instead of  `dimensions`.
- What you're doing with `rowCount` and `hasExtraRow` is a common operation with a classic solution (and further, a specialisation for cases like this one involving power of 2):
  - If we separate out the `* bitsPerPixel` and just apply that to `x_coord` at the beginning, we're left with an integer divide and round up.
  - The classic solution to this problem is `(x + divisor - 1) / divisor`, where the `+ divisor - 1` ensures the result will be 1 greater if `x % divisor != 0`, but has no effect if `x` is an exact multiple of `divisor`.
  - With powers of two we can further specialise this by replacing the `/ divisor` with `>> log2(divisor)`, in this case `>> 5`

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,622 @@
...
+static int32_t _select_bits(PackedByteArray const& data, int32_t rowStart, int32_t firstBit, int32_t bitCount) {
+	int32_t bitmasks[] = {0b1 , 0b11, 0b111, 0b1111, 0b11111, 0b111111, 0b1111111, 0b11111111};
+	int32_t byteIndex = firstBit / 8;
+	int32_t bitInByteIndex = firstBit % 8;
+	if (bitInByteIndex + bitCount > 8){
+		Logger::error("Attempted to select bits outside of a byte.");
+		return 0;
+	}
+	int32_t byte = _reverse_bits(data[rowStart + byteIndex]);
+	int32_t selected = (byte >> bitInByteIndex) & bitmasks[bitCount-1];
```
Hop311, line 231:
```suggestion
static constexpr int32_t _select_bits(PackedByteArray const& data, int32_t rowStart, int32_t firstBit, int32_t bitCount) {
	int32_t byteIndex = firstBit >> 3;
	int32_t bitInByteIndex = firstBit & 0b111;
	if (bitInByteIndex + bitCount > 8) {
		Logger::error("Attempted to select bits outside of a byte.");
		return 0;
	}
	int32_t byte = _reverse_bits(data[rowStart + byteIndex]);
	int32_t selected = (byte >> bitInByteIndex) & ((1 << bitCount) - 1);
```
- If we were to stick with the `bitmasks` array then it should be `static constexpr`, but its values are equivalent to the equation above so we might as well just use it directly and avoid the array in the first place.
- We can replace `/ 8` and `% 8` with bitwise equivalents: `>> 3` and `& 0b111`.
- Also it seems we can actually make this `constexpr` as it only uses the `PackedByteArray` as a `const&` without actually constructing or destructing it or using any of its other functions, although I'm not sure why using its `[]` operator isn't causing any issues (maybe it's treated as a function call to a non-constexpr part and the rest of the function around it is `constexpr`, similar to how `Logger` calls are allowed?)

I believe you could make even more functions `constexpr` by replacing uses of `Vector2i` in arguments with separate `int32_t x, int32_t y` arguments (or just one if you don't need both), and maybe also replacing `PackedByteArray const&` with `uint8_t const*`, got using `PackedByteArray`'s `ptr()` function (I do wonder if we should be testing against the array's length, and similarly checking the length of arrays got from files to make sure they're as long as we expect. It's fine if the code is guarunteed not to look outside of the array's bounds, but if there's a possibility, even just from incorrectly formatted files, then we should add some checks to avoid potential crashes)

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,622 @@
...
+	int32_t r = imageData[offset + (i*4) + 2];
+	int32_t a = imageData[offset + (i*4) + 3] * opaque;
+
+	pixel.append(r);
+	pixel.append(g);
+	pixel.append(b);
+	pixel.append(a);
+
+	return pixel;
+}
```
Hop311, line 303: For `_pixel_palette_lookup`, `_read_24bit_pixel` and `_read_32bit_pixel`, rather than creating a temporary `PackedByteArray` for every single pixel (each one of which requires an individual dynamic memory allocation), why not just pass the main `pixelData` array as a reference and append the bytes to it directly?
Furthermore, it would be even better if we didn't `append`/`append_array` every byte/pixel (which requires a re-allocation every time the array runs out of capacity), but instead `resize`'d the array to the final size at the beginning and then set each byte by index. You could even do this using a mutable pointer (`uint8_t*`, got from the `PackedByteArray` using `ptrw()`), so something like:
```cpp
uint8_t const* src = source_array.ptr(); // source data pointer
uint8_t* dest = destination_array.ptrw(); // destination data pointer

// This sets the byte pointed to by dest to the one pointed to by src and then moves both src and dest along to the next bytes in their repsective arrays.
*dest++ = *src++;

// If a function takes `uint8_t [const]*&` arguments then the original pointer variables given to them will also be moved along, meaning we don't have to do that manually on the caller's end.
// We don't have to move the pointers along immediately, e.g. we could do:
*dest++ = src[2];
*dest++ = src[1];
*dest++ = src[0];
*dest++ = src[3] * opaque;
src += 4;
// This is based on _read_32bit_pixel, copying a BGRA source pixel to a RGBA destination, with both pointers moved along by 4 by the end.
```

Hop311 (commented):
I've left a bunch of comments, the most significant being about handling pixels without creating loads of intermediate `PackedByteArray`s.
Lemme know what you come up with, tbh we're nearing the point where I should just approve this (maybe after Spartan taking a quick look too) and leave further optimisations for the future, as this is working fine but just has so much code that it's easy to keep finding things to tweak

extension/src/openvic-extension/singletons/CursorSingleton.cpp
Nemrav: trying to set PackedByteArray using a pointer (ie. `*arr.ptrw() = ...`) didn't seem to work for me. Just passing in a reference to the output PackedByteArray and using `[]` lets us do this just fine though.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
Nemrav: The [] operator does seem to make the build fail on linux.

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,617 @@
...
+	return value;
+}
+
+static constexpr int32_t _get_row_start(int32_t x_coord, int32_t y_coord, int32_t bitsPerPixel) {
+	x_coord *= bitsPerPixel;
+	int32_t rowCount = (x_coord + 31) >> 5;
+	return rowCount * y_coord * 4; // 4 bytes per row * rows down
+}
+
+//can't be constexpr due to the [] operator on data on Linux
```
Hop311, line 223: Remove this comment

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,617 @@
...
+					}/*
+					else if (bitsPerPixel == 16) { //TODO
+						//Unsupported, error
+					}*/
+					else if (bitsPerPixel == 24) {
+						//Support Questionable, based on 1 example on the internet as opposed to the actual spec
+						_read_24bit_pixel(
+							imageData, pixelData, i, offset, transparent
+						);
+						Logger::info("pixel ",i);
```
Hop311, line 403: Remove this log message

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,617 @@
...
+
+static String to_define_file_name(String const& path) {
+	static const String backslash = "\\";
+	static const String forwardslash = "/";
+	static const String cursor_directory_forwardslash = Utilities::std_to_godot_string(cursor_directory) + forwardslash;
+	static const String dot = ".";
+	return path.replace(backslash, forwardslash).get_slice(cursor_directory_forwardslash, 1).get_slice(dot, 0);
+}
+
+godot::Error CursorSingleton::load_cursors() {
```
Hop311, line 148: You can remove all uses of `godot::` from this source file

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
...
+	var is_animated : bool = false
+	var sequence : PackedInt32Array = [0]
+	var timings : PackedFloat32Array = [1.0]
+	
+	#Cursor state
+	var currentFrame : int = 0
+	var timeToFrame : float = 1.0
+	
+	func _init(nameIn:String) -> void:
+		self.cursor_name = nameIn
```
Hop311, line 19: Not a big deal, but I believe you can remove all uses from `self.` so long as there aren't any variables with conflicting names, e.g. function arguments

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
+extends Node
+
+class CompatCursor:
+	#cursor properties
+	var cursor_name : String
```
Hop311, line 5:
```suggestion
	var cursor_name : StringName
```
Also change the argument type in `_init` and `set_compat_cursor` to `StringName`

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
...
+	func generate_new_res(base_res_index:int, resolution:Vector2) -> void:
+		# resolution wasn't in among the default, need to generate it ourselves
+		CursorSingleton.generate_resolution(cursor_name,base_res_index,resolution)
+
+	#only bother with this if the cursor is animated
+	func _process_cursor(delta:float, shape:Input.CursorShape=Input.CURSOR_ARROW) -> void:
+		timeToFrame -= delta
+		if(timeToFrame <= 0):
+			currentFrame = (currentFrame + 1) % len(sequence)
+			timeToFrame = timings[self.sequence[currentFrame]]
```
Hop311, line 65: Maybe this should be `+=` rather than `=` so we don't ignore whatever portion of `delta` is still left in `timeToFrame`? I guess if there's significant lag then the next couple calls of `_process_cursor` will all trigger a frame change, but that should catch up quickly (we could change the `if` to a `while` and ensure we move to the right frame straight away, maybe holding off on calling `set_hardware_cursor` until the end (assuming we went through the loop at least once) to avoid redundant calls, but tbh this might be overkill for a simple cursor animation function).

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
...
+
+
+#TODO: This is set on game start, but we probably want this to be a video setting
+var preferred_res : Vector2 = Vector2(32,32)
+
+
+var activeCursor : CompatCursor
+var activeShape : Input.CursorShape
+
+#Shape > Cursor dictionnaries
```
Hop311, line 81:
```suggestion
#Shape > Cursor dictionaries
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
...
+	if mouseOverWindow:
+		var mouseShape : Input.CursorShape = Input.get_current_cursor_shape()
+
+		for shape in currentCursors.keys():
+			if currentCursors[shape] != queuedCursors[shape] and queuedCursors[shape] != null:
+				currentCursors[shape] = queuedCursors[shape]
+				currentCursors[shape].set_hardware_cursor(0,shape)
+			elif currentCursors[shape] != queuedCursors[shape]:
+				currentCursors[shape] = queuedCursors[shape]
+				Input.set_custom_mouse_cursor(null,shape)
```
Hop311, line 115:
```suggestion
			if currentCursors[shape] != queuedCursors[shape]:
				currentCursors[shape] = queuedCursors[shape]
				if currentCursors[shape] != null:
					currentCursors[shape].set_hardware_cursor(0, shape)
				else:
					Input.set_custom_mouse_cursor(null, shape)
```
We can avoid repeating `currentCursors[shape] != queuedCursors[shape]` and also only write `currentCursors[shape] = queuedCursors[shape]` once

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,173 @@
...
+			#Current mouse type changed, need to make sure that if the cursor of this new type
+			# is animated, we are providing its frames instead of the frames of the previous active cursor
+			activeShape = mouseShape
+			if mouseShape in currentCursors and currentCursors[mouseShape] != null:
+				activeCursor = currentCursors[mouseShape]
+				activeCursor.reset()
+				activeCursor.set_hardware_cursor(0,activeShape)
+			else:
+				activeCursor = null
+				Input.set_custom_mouse_cursor(null,mouseShape)
```
Hop311, line 129:
```suggestion
			activeCursor = currentCursors.get(activeShape, null)
			if activeCursor != null:
				activeCursor.reset()
				activeCursor.set_hardware_cursor(0, activeShape)
			else:
				Input.set_custom_mouse_cursor(null, activeShape)
```
We can use `currentCursors.get(activeShape, null)` to get the active shape's cursor if it exists and fallback to `null` otherwise, simplifying the logic and reducing the amount of different code in each branch. Also since `activeShape = mouseShape`, it's cleaner to stick to using just one of them after that point

Hop311 (commented):
Great work, I've left some simple GDScript comments, once they're addressed I'm happy to get this merged :)

Hop311 (dismissed):
Another absolutely phenomenal PR, well done Nemrav 🎉 Approved, just with a couple final non-blocking comments:
- I don't see the spinning compass mouse on the loading screen, but that's probably because everything gets loaded in a single call to the SIM repo/dataloader, so we'll need to switch to a staged approach in a later PR to have access to custom cusors before the game finishes loading
- Brick mentioned some remaining Mac issues, up to you and him how to deal with those

Nemrav:
>     * I don't see the spinning compass mouse on the loading screen, but that's probably because everything gets loaded in a single call to the SIM repo/dataloader, so we'll need to switch to a staged approach in a later PR to have access to custom cusors before the game finishes loading

This seems to be a quirk of godot. To get it to appear, you need to initial logo animation away to get the proper loading screen which has the cursor mode marker as "busy" (this is expected). Then, you need to move the mouse so godot registers that it needs to change the cursor to match what the control nodes specify (not what we want). We could get around this at a later date by manually changing the `Arrow_cursor` to "busy" when the stop watch is needed as a work-around.

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+var loaded_cursors : Dictionary = {}
+
+func load_cursors() -> void:
+	CursorSingleton.load_cursors()
+	for cursor_name : StringName in CursorSingleton.cursor_names:
+		var cursor : CompatCursor = CompatCursor.new(cursor_name)
+		cursor.set_resolution(preferred_res)
+		loaded_cursors[cursor_name] = cursor
+
+var mouseOverWindow : bool = false
```
Spartan322, line 101: Doesn't follow style, should be `is_mouse_over_window`.

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+
+#Handle queued cursor changes and cursor animations
+func _process(delta) -> void:
+	#only attempt to update the mouse when this wont crash anything
+	if mouseOverWindow:
+		var mouseShape : Input.CursorShape = Input.get_current_cursor_shape()
```
Spartan322, line 107:
```suggestion
		var mouse_shape : Input.CursorShape = Input.get_current_cursor_shape()
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+		var hotspot : Vector2 = hotspots[sequence[frame]]
+		Input.set_custom_mouse_cursor(texture,shape,hotspot)
+
+
+#TODO: This is set on game start, but we probably want this to be a video setting
+var preferred_res : Vector2 = Vector2(32,32)
+
+
+var activeCursor : CompatCursor
+var activeShape : Input.CursorShape
```
Spartan322, line 79:
```suggestion
var active_cursor : CompatCursor
var active_shape : Input.CursorShape
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+var currentCursors : Dictionary = {
+	Input.CURSOR_ARROW:null,
+	Input.CURSOR_BUSY:null,
+	Input.CURSOR_IBEAM:null
+}
+var queuedCursors : Dictionary = {
+	Input.CURSOR_ARROW:null,
+	Input.CURSOR_BUSY:null,
+	Input.CURSOR_IBEAM:null
+}
```
Spartan322, line 91:
```suggestion
var current_cursors : Dictionary = {
	Input.CURSOR_ARROW:null,
	Input.CURSOR_BUSY:null,
	Input.CURSOR_IBEAM:null
}
var queued_cursors : Dictionary = {
	Input.CURSOR_ARROW:null,
	Input.CURSOR_BUSY:null,
	Input.CURSOR_IBEAM:null
}
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+	var resolutions : PackedVector2Array
+	var frames : Array[ImageTexture]
+	var hotspots : PackedVector2Array
+	var is_animated : bool = false
+	var sequence : PackedInt32Array = [0]
+	var timings : PackedFloat32Array = [1.0]
+	
+	#Cursor state
+	var currentFrame : int = 0
+	var timeToFrame : float = 1.0
```
Spartan322, line 16:
```suggestion
	var current_frame : int = 0
	var time_to_frame : float = 1.0
```

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,89 @@
...
+
+	private:
+		//.cur files use all but the last 2 properties, rest are for .ani
+		//images is an array of ImageTexture, just godot doesn't like typed arrays with image texture for some reason
+		struct cursor_asset_t {
+			std::vector<godot::PackedVector2Array> hotspots;
+			std::vector<godot::TypedArray<godot::ImageTexture>> images;
+			godot::PackedVector2Array resolutions;
+			int32_t animationLength; //1 for static cursors
+			std::optional<godot::PackedFloat32Array> displayRates;
```
Spartan322, line 49:
```suggestion
			std::optional<godot::PackedFloat32Array> display_rates;
```

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,89 @@
...
+namespace OpenVic {
+
+	class CursorSingleton : public godot::Object {
+		
+		/*
+		This singleton only handles the dataloading for the cursors.
+		The functionality for actually setting one of these as a cursor
+		and animating animated cursors is done by the CursorManager autoloaded
+		gd script.
+		*/
```
Spartan322, line 28: Documentation should be in `CursorSingleton.xml` following https://docs.godotengine.org/en/latest/contributing/documentation/index.html

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,89 @@
...
+	private:
+		//primary loading functions
+		bool _load_cursor_ani(godot::StringName const& name, godot::String const& path);
+		bool _load_cursor_cur(godot::StringName const& name, godot::String const& path);
+
+	public:
+		godot::Error load_cursors();
+		godot::TypedArray<godot::StringName> get_cursor_names() const;
+		
+		//getters for the different cursor properties
```
Spartan322, line 77:
```suggestion
```

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,89 @@
...
+		godot::TypedArray<godot::ImageTexture> get_frames(godot::StringName const& name, int32_t res_index = 0) const;
+		godot::PackedVector2Array get_hotspots(godot::StringName const& name, int32_t res_index = 0) const;
+		int32_t get_animationLength(godot::StringName const& name) const;
+		godot::PackedVector2Array get_resolutions(godot::StringName const& name) const;
+		godot::PackedFloat32Array get_displayRates(godot::StringName const& name) const;
+		godot::PackedInt32Array get_sequence(godot::StringName const& name) const;
+
+		void generate_resolution(godot::StringName const& name, int32_t base_res_index, godot::Vector2 target_res);
+
```
Spartan322, line 86:
```suggestion
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+				highest_res_index = i
+		generate_new_res(highest_res_index,resolution)
+		
+		resolutions = CursorSingleton.get_resolutions(cursor_name)
+		frames = CursorSingleton.get_frames(cursor_name,len(resolutions)-1)
+		hotspots = CursorSingleton.get_hotspots(cursor_name,len(resolutions)-1)
+		
+		assert(len(frames) != 0)
+
+	func generate_new_res(base_res_index:int, resolution:Vector2) -> void:
```
Spartan322, line 56:
```suggestion
	func generate_new_resolution(base_res_index:int, resolution:Vector2) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+				activeCursor.set_hardware_cursor(0, activeShape)
+			else:
+				Input.set_custom_mouse_cursor(null, activeShape)
+
+		#if we didnt change cursors and are animated, do an update
+		elif activeCursor != null and activeCursor.is_animated:
+			activeCursor._process_cursor(delta,activeShape)
+
+
+func set_prefered_res(res_in:Vector2) -> void:
```
Spartan322, line 135:
```suggestion
func set_prefered_resolution(res_in:Vector2) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,172 @@
...
+	set_compat_cursor(&"busy", Input.CURSOR_BUSY)
+
+# To safely change the mouse cursor, the mouse must be over the window
+# this function helps ensure we do it safely
+func _notification(what) -> void:
+	match(what):
+		NOTIFICATION_WM_MOUSE_ENTER:
+			mouseOverWindow = true
+		NOTIFICATION_WM_MOUSE_EXIT:
+			mouseOverWindow = false
```
Spartan322, line 172: Probably should be connected Window.mouse_entered and Window.mouse_exited in a `_ready()` method, for example:
```gdscript
func _ready() -> void:
  var window : Window = get_window()
  window.mouse_entered.connect(func() -> void: is_mouse_over_window = true)
  window.mouse_exited.connect(func() -> void: is_mouse_over_window = false)
```

Nemrav: If you attempt to call an Input hardware cursor switch while the mouse isn't over the window, godot just crashes.

Spartan322: Has that crash been reported to Godot?

Spartan322: Also have you tested this on 4.3?

Nemrav: Things seem to work fine with the mouse over window check removed. I guess the problem, whatever it was, was fixed with 4.3

Nemrav: Regardless, the mouse over window check was unnecessary and removed. Things should still be working fine.

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+	<tutorials>
+	</tutorials>
+	<methods>
+		<method name="generate_resolution">
+			<return type="void" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="base_resolution_index" type="int" default="0" />
+			<param index="2" name="target_resolution" type="Vector2" default="Vector2(64, 64)" />
+			<description>
+				Takes the cursor image at `base_resolution_index` then scales it to `target_resolution` then saves it to the image as an extra resolution option.
```
Spartan322, line 17:
```suggestion
				Takes the cursor image at [param base_resolution_index] and scales it to [param target_resolution] then saves it to the image as an extra resolution option.
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<class name="CursorSingleton" inherits="Object" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/godotengine/godot/master/doc/class.xsd">
+	<brief_description>
+	</brief_description>
+	<description>
+		This singleton handles the dataloading and access to the windows .cur and .ani cursors found in the /cursors folder. The functionality for setting and animating the cursors is done by the CursorManager gd script.
```
Spartan322, line 6:
```suggestion
		This singleton handles the dataloading and access to the windows .cur and .ani cursors found in the [code]/cursors[/code] folder. The functionality for setting and animating the cursors is done by the CursorManager gd script.
```

Nemrav: whoops, there'll be more of that, I'll go through the doc again now.

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			<param index="2" name="target_resolution" type="Vector2" default="Vector2(64, 64)" />
+			<description>
+				Takes the cursor image at `base_resolution_index` then scales it to `target_resolution` then saves it to the image as an extra resolution option.
+			</description>
+		</method>
+		<method name="get_animationLength" qualifiers="const">
+			<return type="int" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns the length of the sequence for the cursor `cursor_name`. This is always `1` for `.cur` cursors. This will be greater than or equal to the number of animation frames since the animation sequence can use a given frame more than once.
```
Spartan322, line 24:
```suggestion
				Returns the length of the sequence for the cursor [param cursor_name]. This will be greater than or equal to the number of animation frames.
				[code].cur[/code] cursors always return [code]1[/code]. 
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+		<method name="get_displayRates" qualifiers="const">
+			<return type="PackedFloat32Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns an array containing how long each frame should last for an animated cursor `cursor_name`. The size of this array will be the same as the size of the sequence array.
```
Spartan322, line 31:
```suggestion
				Returns an array containing how long each frame should last for an animated cursor [param cursor_name].
				The size of the array will be the same as the size of the sequence array.
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+		<method name="get_frames" qualifiers="const">
+			<return type="ImageTexture[]" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="resolution_index" type="int" default="0" />
+			<description>
+				Returns an array of animation frames given a `cursor_name` and a `resolution_index`.
```
Spartan322, line 39:
```suggestion
				Returns an array of animation frames given a [param cursor_name] and [param resolution_index].
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+		<method name="get_hotspots" qualifiers="const">
+			<return type="PackedVector2Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<param index="1" name="resolution_index" type="int" default="0" />
+			<description>
+				Returns an array of cursor click positions given a `cursor_name` and `resolution_index`.
```
Spartan322, line 47:
```suggestion
				Returns an array of cursor click positions given a [param cursor_name] and [param resolution_index].
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns an array of all the image resolutions contained in the cursor.
+			</description>
+		</method>
+		<method name="get_sequence" qualifiers="const">
+			<return type="PackedInt32Array" />
+			<param index="0" name="cursor_name" type="StringName" default="&quot;normal&quot;" />
+			<description>
+				Returns a list of frame indices used to produce an animation. Displaying the frames in the order specified by this sequence will produce the intended animation.
```
Spartan322, line 61:
```suggestion
				Returns a list of frame indices used to produce an animation.
				Displaying the frames in the order specified by this sequence will produce the intended animation.
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+		<method name="load_cursors">
+			<return type="int" enum="Error" />
+			<description>
+				Loads the cursors from the `/cursors` folder. This function must be called before any other cursor related function.
```
Spartan322, line 67:
```suggestion
				Loads the cursors from the [code]/cursors[/code] folder. This function must be called before any other cursor related function.
```

extension/doc_classes/CursorSingleton.xml
```diff
@@ -0,0 +1,76 @@
...
+			</description>
+		</method>
+	</methods>
+	<members>
+		<member name="cursor_names" type="StringName[]" setter="" getter="get_cursor_names" default="[]">
+			A list of cursor file names. These names are the possible valid `cursor_name` parameters for the getter functions.
```
Spartan322, line 73:
```suggestion
			A list of cursor file names.
```

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,78 @@
...
+	public:
+		godot::Error load_cursors();
+		godot::TypedArray<godot::StringName> get_cursor_names() const;
+		
+		godot::TypedArray<godot::ImageTexture> get_frames(godot::StringName const& name, int32_t res_index = 0) const;
+		godot::PackedVector2Array get_hotspots(godot::StringName const& name, int32_t res_index = 0) const;
+		int32_t get_animationLength(godot::StringName const& name) const;
+		godot::PackedVector2Array get_resolutions(godot::StringName const& name) const;
+		godot::PackedFloat32Array get_displayRates(godot::StringName const& name) const;
```
Spartan322, line 72:
```suggestion
		godot::PackedFloat32Array get_display_rates(godot::StringName const& name) const;
```

extension/src/openvic-extension/singletons/CursorSingleton.hpp
```diff
@@ -0,0 +1,78 @@
...
+		bool _load_cursor_ani(godot::StringName const& name, godot::String const& path);
+		bool _load_cursor_cur(godot::StringName const& name, godot::String const& path);
+
+	public:
+		godot::Error load_cursors();
+		godot::TypedArray<godot::StringName> get_cursor_names() const;
+		
+		godot::TypedArray<godot::ImageTexture> get_frames(godot::StringName const& name, int32_t res_index = 0) const;
+		godot::PackedVector2Array get_hotspots(godot::StringName const& name, int32_t res_index = 0) const;
+		int32_t get_animationLength(godot::StringName const& name) const;
```
Spartan322, line 70:
```suggestion
		int32_t get_animation_length(godot::StringName const& name) const;
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,615 @@
...
+void CursorSingleton::_bind_methods() {
+	OV_BIND_METHOD(CursorSingleton::load_cursors);
+	OV_BIND_METHOD(CursorSingleton::get_frames,{"cursor_name","resolution_index"},"normal",0);
+	OV_BIND_METHOD(CursorSingleton::get_hotspots,{"cursor_name","resolution_index"},"normal",0);
+	OV_BIND_METHOD(CursorSingleton::get_animationLength,{"cursor_name"},"normal");
+	OV_BIND_METHOD(CursorSingleton::get_displayRates,{"cursor_name"},"normal");
+	OV_BIND_METHOD(CursorSingleton::get_sequence,{"cursor_name"},"normal");
+	OV_BIND_METHOD(CursorSingleton::get_resolutions,{"cursor_name"},"normal");
+	OV_BIND_METHOD(CursorSingleton::generate_resolution,{"cursor_name","base_resolution_index", "target_resolution"},"normal",0,Vector2(64,64));
+	OV_BIND_METHOD(CursorSingleton::get_cursor_names);
```
Spartan322, line 33:
```suggestion
	OV_BIND_METHOD(CursorSingleton::load_cursors);
	OV_BIND_METHOD(CursorSingleton::get_frames, {"cursor_name", "resolution_index"}, "normal", 0);
	OV_BIND_METHOD(CursorSingleton::get_hotspots, {"cursor_name", "resolution_index"}, "normal", 0);
	OV_BIND_METHOD(CursorSingleton::get_animation_length, {"cursor_name"}, "normal");
	OV_BIND_METHOD(CursorSingleton::get_display_rates, {"cursor_name"}, "normal");
	OV_BIND_METHOD(CursorSingleton::get_sequence, {"cursor_name"}, "normal");
	OV_BIND_METHOD(CursorSingleton::get_resolutions, {"cursor_name"}, "normal");
	OV_BIND_METHOD(CursorSingleton::generate_resolution, {"cursor_name", "base_resolution_index", "target_resolution"}, "normal", 0, Vector2(64,64));
	OV_BIND_METHOD(CursorSingleton::get_cursor_names);
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,615 @@
...
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	std::vector<PackedVector2Array> const& hotspots = it->second.hotspots;
+	ERR_FAIL_INDEX_V_MSG(res_index, hotspots.size(), {}, vformat("Invalid hotspot index for cursor \"%s\": %d", name, res_index));
+
+	return hotspots[res_index];
+}
+
+int32_t CursorSingleton::get_animationLength(StringName const& name) const {
```
Spartan322, line 80:
```suggestion
int32_t CursorSingleton::get_animation_length(StringName const& name) const {
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,615 @@
...
+}
+
+PackedVector2Array CursorSingleton::get_resolutions(StringName const& name) const {
+	const cursor_map_t::const_iterator it = cursors.find(name);
+	ERR_FAIL_COND_V_MSG(it == cursors.end(), {}, vformat("Cursor \"%s\" not found", name));
+
+	return it->second.resolutions;
+}
+
+PackedFloat32Array CursorSingleton::get_displayRates(StringName const& name) const {
```
Spartan322, line 94:
```suggestion
PackedFloat32Array CursorSingleton::get_display_rates(StringName const& name) const {
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,615 @@
...
+	}
+
+	cursor.images.push_back(new_frameset);
+	cursor.hotspots.push_back(new_hotspots);
+	cursor.resolutions.push_back(target_res);
+}
+
+static constexpr std::string_view cursor_directory = "gfx/cursors";
+
+static String to_define_file_name(String const& path) {
```
Spartan322, line 140:
```suggestion
static String _to_define_file_name(String const& path) {
```

extension/src/openvic-extension/singletons/CursorSingleton.cpp
```diff
@@ -0,0 +1,615 @@
...
+	pixel_data[(i*4) + 2] = imageData[offset + (i*4) + 0]; //b
+	pixel_data[(i*4) + 3] = imageData[offset + (i*4) + 3] * opaque; //a
+}
+
+//used to load a .cur file from a file (could be the a whole .cur file, or a .cur within a .ani file)
+static CursorSingleton::image_hotspot_pair_asset_t _load_pair(Ref<FileAccess> const& file) {
+	CursorSingleton::image_hotspot_pair_asset_t pairs = {};
+
+	//.cur's within .anis won't start of the beginning of the file, so save where they start
+	int32_t baseOffset = file->get_position();
```
Spartan322, line 309:
```suggestion
	int32_t base_offset = file->get_position();
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+
+	#only bother with this if the cursor is animated
+	func _process_cursor(delta:float, shape:Input.CursorShape=Input.CURSOR_ARROW) -> void:
+		time_to_frame -= delta
+		if(time_to_frame <= 0):
+			current_frame = (current_frame + 1) % len(sequence)
+			time_to_frame += timings[sequence[current_frame]] 
+			set_hardware_cursor(current_frame, shape)
+
+	func set_hardware_cursor(frame:int=0, shape:Input.CursorShape=Input.CURSOR_ARROW) -> void:
```
Spartan322, line 68:
```suggestion
	func set_hardware_cursor(frame : int = 0, shape : Input.CursorShape = Input.CURSOR_ARROW) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+		hotspots = CursorSingleton.get_hotspots(cursor_name,len(resolutions)-1)
+		
+		assert(len(frames) != 0)
+
+	func generate_new_resolution(base_res_index:int, resolution:Vector2) -> void:
+		# resolution wasn't in among the default, need to generate it ourselves
+		CursorSingleton.generate_resolution(cursor_name,base_res_index,resolution)
+
+	#only bother with this if the cursor is animated
+	func _process_cursor(delta:float, shape:Input.CursorShape=Input.CURSOR_ARROW) -> void:
```
Spartan322, line 61:
```suggestion
	func _process_cursor(delta : float, shape : Input.CursorShape = Input.CURSOR_ARROW) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+				highest_res_index = i
+		generate_new_resolution(highest_res_index,resolution)
+		
+		resolutions = CursorSingleton.get_resolutions(cursor_name)
+		frames = CursorSingleton.get_frames(cursor_name,len(resolutions)-1)
+		hotspots = CursorSingleton.get_hotspots(cursor_name,len(resolutions)-1)
+		
+		assert(len(frames) != 0)
+
+	func generate_new_resolution(base_res_index:int, resolution:Vector2) -> void:
```
Spartan322, line 56:
```suggestion
	func generate_new_resolution(base_res_index : int, resolution : Vector2) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+	var hotspots : PackedVector2Array
+	var is_animated : bool = false
+	var sequence : PackedInt32Array = [0]
+	var timings : PackedFloat32Array = [1.0]
+	
+	#Cursor state
+	var current_frame : int = 0
+	var time_to_frame : float = 1.0
+	
+	func _init(nameIn:StringName) -> void:
```
Spartan322, line 18:
```suggestion
	func _init(name_in : StringName) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+			active_cursor.set_hardware_cursor(0, active_shape)
+		else:
+			Input.set_custom_mouse_cursor(null, active_shape)
+
+	#if we didnt change cursors and are animated, do an update
+	elif active_cursor != null and active_cursor.is_animated:
+		active_cursor._process_cursor(delta,active_shape)
+
+
+func set_prefered_resolution(res_in:Vector2) -> void:
```
Spartan322, line 131:
```suggestion
func set_prefered_resolution(res_in : Vector2) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+	preferred_res = res_in
+
+func set_compat_cursor(cursor_name:StringName, shape:Input.CursorShape=Input.CURSOR_ARROW) -> void:
```
Spartan322, line 134:
```suggestion
func set_compat_cursor(cursor_name : StringName, shape : Input.CursorShape = Input.CURSOR_ARROW) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+func load_cursors() -> void:
+	CursorSingleton.load_cursors()
+	for cursor_name : StringName in CursorSingleton.cursor_names:
+		var cursor : CompatCursor = CompatCursor.new(cursor_name)
+		cursor.set_resolution(preferred_res)
+		loaded_cursors[cursor_name] = cursor
+
+
+#Handle queued cursor changes and cursor animations
+func _process(delta) -> void:
```
Spartan322, line 102:
```suggestion
func _process(delta : float) -> void:
```

game/src/Game/Autoload/CursorManager.gd
```diff
@@ -0,0 +1,160 @@
...
+# to indicate when window is busy, normal, doing a drag-select, etc. 
+# You can set this per Control Node under Mouse > Default Cursor Shape 
+
+# set_compat_cursor makes the named vic2 cursor the presently active
+# one for the shape it is currently associated with. By default a cursor
+# is associated with Input.CURSOR_ARROW, but you can override this with the second
+# argument. Use set_compat_cursor as you find it used here in initial_cursor_setup().
+
+func initial_cursor_setup() -> void:
+	set_prefered_resolution(Vector2(48,48))
```
Spartan322, line 154:
```suggestion
	set_preferred_resolution(Vector2(32, 32))
```

game/src/Game/GameSession/GameSessionMenu.gd
```diff
@@ -49,6 +49,7 @@ func show_save_dialog_button() -> void:
 func _on_main_menu_confirmed() -> void:
 	SaveManager.current_session_tag = ""
 	SaveManager.current_save = null
+	CursorManager.set_compat_cursor(&"", Input.CURSOR_IBEAM)
```
Hop311, line 52: Why do we need to set the cursor to `&""` here?

Hop311 (approved):
LGTM on Windows, and we can fix any Mac issues in a future PR. It can be hard to catch the busy icon when the game is loading, often I only see the normal one but I think that's mostly down to timing issues between the game window and the dataloader.

game/src/Game/GameSession/GameSessionMenu.gd
Nemrav: This is to reset the cursor back to null (ie. OS default cursors for the IBEAM on the main menu). null can't be passed to a StringName, so this is the next best thing.

PR #253, opened 2024-08-26
Tooltips

Hop311:
Adds tooltip support for:
- buttons (just works)
- labels and icons (mouse filter must be changed from IGNORE to PASS before adding a tooltip, and the node should be reparented if it would block clicks to a sibling button)
- TODO: pie charts, progress bars, maybe more (line graphs too once they're implemented)

Various tooltips are implemented as examples, some using placeholder text.


Spartan322 (dismissed):
Looks really good, works great on Linux

PR #256, opened 2024-08-31
Document industry score

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -147,21 +147,36 @@ POPs:
...
-              2*: Hardcoded value.
\ No newline at end of file
+              2 *: Hardcoded value.
+Scores:
+  Industrial power:
+    formula: sum of investment scores + sum of state factory scores
+    investment score: money_invested * defines.country.INVESTMENT_SCORE_FACTOR / 100
+    state factory score:
+      if n_workers > 0:
+        formula: sum_factory_levels_in_state * max(0.2, min(4, floor(n_workers_in_state / 100) * 400 / sum_workforce_in_state))
```
Hop311, line 174: When you have lower and upper bounds for a value you could use `clamp(value, low, high)` instead of `max(low, min(high, value))`. It's a common function in languages like [C++](https://en.cppreference.com/w/cpp/algorithm/clamp) and especially in shader languages, and IMO it's easier to read.

Hop311 (approved):
Excellent research, thank you 🙏 I'll be using this in my upcoming score/ranking PR

docs/simulation/calculations.yaml
wvpm: I am aware of `clamp`. I avoid using it as there are different orders for the arguments per language.
css for example uses `clamp(min, val, max)`. `min` and `max` are unambiguous.

PR #257, opened 2024-09-01
Military score calculation

wvpm:


docs/simulation/calculations.yaml
```diff
@@ -178,5 +178,19 @@ Scores:
...
+      formula: disarmed penalty * regular army size * supply consumption * sum unit stats / (7 * (1 + n_unit_types))
+      disarmed penalty:
+        if is_disarmed: defines.diplomacy.DISARMAMENT_ARMY_HIT
+        else: 1
+      regular army size: min(4 * number of deployed non-mobilised regiments, possible regiment count)
+      supply consumption: From army tab, not unit types.
+      sum unit stats: Sum of ((attack + defence + land_attack_modifier + land_defense_modifier) * discipline) for each land unit type (including locked).
+      n_unit_types: Number of land unit types (including locked).
+      7*: Hardcoded value.
+    capital ship score: Sum of ((hull + naval_defense_modifier) * (gun_power + naval_attack_modifier)) for each capital ship instance.
```
Hop311, line 195: Do the attack/defense modifiers include the base `1`, or are they just the extra modifiers? Either way, in many cases the attack/defense modifier is tiny compared to `gun_power`/`hull`, going as far as dreadnoughts having 60 base `gun_power` and 80 base `hull`. It would make more sense to me for the `*` and `+` to be swapped in this equation, but I guess it could just be paradox maths 🙃

Hop311 (dismissed):

docs/simulation/calculations.yaml
wvpm: Both attack and defense modifiers are 0 by default.
The naval modifiers show a value of 1 as +100% (which is false, it's just a flat +1).
The land modifiers don't show a value of 1 and instead display a value of 2 as +100%.

Hop311 (approved):

PR #266, opened 2024-09-08
Ranking + army topbar info

Hop311:


extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -31,12 +31,18 @@ String const& GUILabel::get_substitution_marker() {
 	return SUBSTITUTION_MARKER;
 }
 
+String const& GUILabel::get_flag_marker() {
```
wvpm, line 34: (Probably me lacking context)
Why is this a method instead of a constant value?

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -452,6 +502,157 @@ int32_t MenuSingleton::get_rgo_owner_pop_icon_index() const {
...
+Dictionary MenuSingleton::get_topbar_info() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, {});
+
+	CountryInstance const* country = game_singleton->get_viewed_country();
+	if (country == nullptr) {
+		return {};
+	}
+
+	Dictionary ret;
```
wvpm, line 516: Why dictionary instead of class/struct?

wvpm (approved):

extension/src/openvic-extension/classes/GUILabel.cpp
Hop311: `static` variables outside of functions are initialised at the very beginning of the program (before the `main` function is called), but `static` variables inside functions are only initialised when the function is first called/the line defining the variable is first reached.
We can't initialise any `String`s or `StringName`s before Godot/GDExtension has run its own setup code, otherwise it causes crashes/breaks the library loading process (our code is compiled as a library that is dynamically loaded by the Godot Editor or the exported Godot program when they start up). So any constants like `FLAG_MARKER` have to be wrapped in a function to delay their initialisation.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: A custom struct would have to be registered with Godot to be usable at the GDScript level. In cases like this that's a lot more code to write for little gain (GDScript looks up class member variables by name in the same way as Dictionary entries are looked up, so if there's any performance difference I'd expect classes to be the slower approach).

PR #271, opened 2024-09-12
Update SIM repo to include resulant modifier calculation

Hop311:
Check out the colonial power tooltip to see the overall modifiers affecting the current country (and right click on the map to select other countries).

### DIFF
--- a/extension/src/openvic-extension/singletons/MenuSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/MenuSingleton.cpp
@@ -3,7 +3,7 @@
 #include <godot_cpp/variant/utility_functions.hpp>
 
 #include <openvic-simulation/GameManager.hpp>
-#include <openvic-simulation/misc/Modifier.hpp>
+#include <openvic-simulation/modifier/Modifier.hpp>
 
 #include "openvic-extension/classes/GFXPieChartTexture.hpp"
 #include "openvic-extension/classes/GUINode.hpp"


Nemrav (commented):
Tested, works and looks good.

Nemrav (approved):

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -698,7 +698,10 @@ Dictionary MenuSingleton::get_topbar_info() const {
 	// TODO - colonial power info
 	ret[colonial_power_available_key] = 0;
 	ret[colonial_power_max_key] = 0;
-	ret[colonial_power_tooltip_key] = String {};
+	ret[colonial_power_tooltip_key] =
+		// TODO - remove these examples eventually
+		// make_rules_tooltip(country->get_rule_set()) + get_tooltip_separator() +
+		make_modifier_effects_tooltip(country->get_modifier_sum().get_value_sum());
```
wvpm, line 704: This is debugging/testing code for modifiers. It does not show colonial power.
So don't merge to master?

Hop311: I was thinking it would be nice for users to be able to see evidence of the modifier calculating system if they check out the game as it currently stands, but it's not that important

wvpm (approved):

PR #273, opened 2024-09-16
Model loader Fixes

Nemrav:
Fixes the following problems:
- Arab_infantry_helmet
- Parts of units being culled when near edge of screen (added cull margin)
- Improper method of determining when to use int16 or float32 quaternions for xsm animations.

Background of Arab_infantry_helmet
Paradox made the helmet models directly on the infantry units they go with. These helmet models are properly separated to their own .xac files so they are normally reuseable attachments. However, the mesh for the arab_infantry_helmet was never removed from the infantry model, so it was appearing twice. So a check was made so that we skip building the helmet's mesh if it is in a file with other mesh chunks (ie. the infantry model)

New method of determining whether to use int16 or float32 for xsm quaternions is to check the chunk version, as opposed to the previous hack of looking in the header chunk's padding for data.

Cull margin was added as a const at the top of the xac loader file, this value can be played with, but I found `= 2` to work best (`= 1` was too small, larger seems unnecessary).

### DIFF
--- a/game/src/Game/Model/XACLoader.gd
+++ b/game/src/Game/Model/XACLoader.gd
@@ -2,6 +2,7 @@ class_name XACLoader
 
 static var unit_shader : ShaderMaterial = preload("res://src/Game/Model/unit_colours_mat.tres")
 const MAX_UNIT_TEXTURES : int = 32 # max number of textures supported by the shader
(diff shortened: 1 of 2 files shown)


Hop311 (approved):
LGTM 👍

PR #275, opened 2024-09-18
Allow GFXPieChartTexture to use any distribution with identifiers and colours rather than just those derived from HasIdentifierAndColour

Hop311:


BrickPi (approved):
LGTM

PR #278, opened 2024-09-25
Document rgo production

wvpm:
Document RGO production

### DIFF
--- a/docs/simulation/calculations.yaml
+++ b/docs/simulation/calculations.yaml
@@ -51,7 +51,44 @@ Economy:
       formula: MGQ * GOLD_TO_CASH_RATE
       MGQ: Money good quanity = total production of goods with money = yes
       GOLD_TO_CASH_RATE: defines.country.GOLD_TO_CASH_RATE
+    Overseas maintenance:
+      See: https://github.com/schombert/Project-Alice/blob/b7889a9c302feed851c039f1a98d73b75be3215a/docs/rules.md#overseas-penalty
   Production:
+    Resource gathering operation:
+      base workforce: defined in common/production_types.txt as workforce
+      rgo size:
+        formula: floor(1.5 * ceil(n_workers_in_state / (base workforce * (1 + terrain rgo size modifier)))
+        n_workers_in_state: Total size of POPs in state where POP type is in employees (common/production_types.txt).
+        note: Calculated when starting a new game.
+      Output:
+        formula: base output * rgo size * (1 + overseas penalty) * throughput from workers * (1 + throughput modifier) * (1 + output from workers) * (1 + output modifier)
+        base output: defined in common/production_types.txt
+        overseas penalty:
+          if province is 'overseas': overseas penalty (see economy > budget > overseas maintenance)
+          else: 0
+        workforce:
+          formula: base workforce * rgo size * (1 + rgo size modifier)
+          base workforce: Defined in common/production_types.txt as workforce.
+          rgo size modifier: sum of rgo size modifiers + sum of farm size modifiers (if farm=yes) + sum of mine size modifiers (if mine=yes)
+        job portion of workforce:
+          formula: employees with job / workforce
+          employees with job: Number of POPs employed with the job. Job being an entry in employees in common/production_types.txt.
+        effect from job:
+          formula:
+            if effect_multiplier == 1: effect_multiplier * employees with job / workforce
+            else: effect_multiplier * min(employees with job / workforce, amount)
+          effect_multiplier: defined in common/production_types.txt as part of employees (default 1)
+          amount: defined in common/production_types.txt (default 1)
+        throughput from workers: sum of effect from job for each job with `effect = throughput`
+        output from workers: sum of effect from job for each job with `effect = output`
+        throughput modifier: sum of throughput modifiers (including from owner job)
+        output modifier: sum of output modifiers (including from owner job)
+        modifier from owner job:
+          formula: effect_multiplier * n_owners_in_state / n_pops_in_state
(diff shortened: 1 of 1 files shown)


docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
       formula: MGQ * GOLD_TO_CASH_RATE
       MGQ: Money good quanity = total production of goods with money = yes
       GOLD_TO_CASH_RATE: defines.country.GOLD_TO_CASH_RATE
+    Overseas maintenance:
+      See: https://github.com/schombert/Project-Alice/blob/main/docs/rules.md#overseas-penalty
```
Hop311, line 55: It would be better for this to have the overseas maintenance explanation copied into text directly here with the link as a reference, rather than only having the link.
Also it's probably better to use a permanent link (`https://github.com/schombert/Project-Alice/blob/b7889a9c302feed851c039f1a98d73b75be3215a/docs/rules.md#overseas-penalty`) so this always refers to the same text, rather than using a link to their `main` branch which could change (e.g. the file gets moved and the link invalidated).

docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
...
   Production:
+    Resource gathering operation:
+      base workforce: defined in common/production_types.txt as workforce
+      rgo size:
+        formula: floor(1.5*ceil(n_workers_in_state / (base workforce * (1 + terrain rgo size modifier)))
```
Hop311, line 60:
```suggestion
        formula: floor(1.5 * ceil(n_workers_in_state / (base workforce * (1 + terrain rgo size modifier)))
```
Consistent spacing would look cleaner

docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
...
+        n_workers_in_state: Total size of POPs in state where POP type is in employees (common/production_types.txt).
+        note: Calculated when starting a new game.
+      Output:
+        formula: base output * rgo size * (1 + overseas penalty) * throughput from workers * (1 + throughput modifier) * (1 + output from workers) * (1 + output modifier)
```
Hop311, line 64: So `throughput from workers` is what makes workers create output (with 0 output when there's 0 throughput causing workers), while `output from workers` boosts output further without generally controlling whether it is 0 or not.

docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
...
+          formula: base workforce * rgo size * (1 + rgo size modifier)
+          base workforce: Defined in common/production_types.txt as workforce.
+          rgo size modifier: sum of rgo size modifiers + sum of farm size modifiers (if farm=yes) + sum of mine size modifiers (if mine=yes)
+        job portion of workforce:
+          formula: employees with job / workforce
+          employees with job: Number of POPs employed with the job. Job being an entry in employees in common/production_types.txt.
+        effect from job:
+          formula:
+            if effect_multiplier == 1: effect_multiplier * employees with job / workforce
+            else: effect_multiplier * min(employees with job / workforce, amount)
```
Hop311, line 79: So `amount` is just ignored if `effect_multiplier == 1`? And if `effect_multiplier != 1` then `amount` acts as a maximum limit on `employees with job / workforce`?
Is this not maybe the other way around, as if `amount == 1` then `min(employees with job / workforce, amount) = employees with job / workforce` because you can't have more employees than workforce?

docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
...
+          effect_multiplier: defined in common/production_types.txt (default 1)
+          amount: defined in common/production_types.txt (default 1)
+        throughput from workers: sum of effect from job for each job with `effect = throughput`
+        output from workers: sum of effect from job for each job with `effect = output`
+        throughput modifier: sum of throughput modifiers (including from owner)
+        output modifier: sum of throughput modifiers (including from owner)
+        modifier from owner:
```
Hop311, line 86:
```suggestion
        throughput modifier: sum of throughput modifiers (including from owners)
        output modifier: sum of output modifiers (including from owners)
        modifier from owners:
```
Change `throughput` to `output`.
Maybe change `owner` to `owners` in these lines, just to make clear this doesn't refer to the country owning the province (which was my initial assumption before reading further along).

docs/simulation/calculations.yaml
```diff
@@ -51,7 +51,44 @@ Economy:
...
+          formula: effect_multiplier * n_owners_in_state / n_pops_in_state
+          effect_multiplier: see throughput from workers
```
Hop311, line 88: Could you make it more clear that `modifier from owners` refers to "owner jobs" and their own `effect` and `effect_multiplier` rather than using employee jobs' `effect`s and `effect_multiplier`s.

Hop311 (commented):
Great work, left a couple comments including one typo.

docs/simulation/calculations.yaml
wvpm: Correct.

docs/simulation/calculations.yaml
wvpm: You can permanently have more employees than workforce for reasons I do not yet fully understand.
Also you can temporarily have more employees than workforce when workforce is decreased via modifiers. (Employees would be fired but that takes days.)

When effect_mutliplier is exactly equal to 1 you can scale beyond max workforce.
Otherwise, having more employees than workforce doesn't offer any benefit.

Hop311 (approved):
Thanks for implementing those changes 👍

PR #296, opened 2024-11-20
Add Map Billboards

Nemrav:
Billboards are used for various mapmodes, and for displaying nation capitals on the map.
Bunch of TODOS:
- Move billboards to own viewport so they don't clip into unit models or occlude them
- hide billboards when zoomed out
- Make billboards correspond to the map mode
- Stop capital billboards and map mode billboards from occluding eachother
- Fix billboard lighting
- Get billboards to display in the proper order if possible (occasionally a billboard will display behind another one which is further back)

### DIFF
--- a/extension/src/openvic-extension/register_types.cpp
+++ b/extension/src/openvic-extension/register_types.cpp
@@ -24,6 +24,7 @@
 #include "openvic-extension/singletons/Checksum.hpp"
 #include "openvic-extension/singletons/GameSingleton.hpp"
 #include "openvic-extension/singletons/LoadLocalisation.hpp"
+#include "openvic-extension/singletons/MapItemSingleton.hpp"
 #include "openvic-extension/singletons/MenuSingleton.hpp"
 #include "openvic-extension/singletons/ModelSingleton.hpp"
 #include "openvic-extension/singletons/SoundSingleton.hpp"
@@ -34,6 +35,7 @@ using namespace OpenVic;
 static Checksum* _checksum_singleton = nullptr;
 static LoadLocalisation* _load_localisation = nullptr;
 static GameSingleton* _game_singleton = nullptr;
+static MapItemSingleton* _map_item_singleton = nullptr;
 static MenuSingleton* _menu_singleton = nullptr;
 static ModelSingleton* _model_singleton = nullptr;
 static AssetManager* _asset_manager_singleton = nullptr;
@@ -55,6 +57,10 @@ void initialize_openvic_types(ModuleInitializationLevel p_level) {
 	ClassDB::register_class<GameSingleton>();
 	_game_singleton = memnew(GameSingleton);
 	Engine::get_singleton()->register_singleton("GameSingleton", GameSingleton::get_singleton());
+
+	ClassDB::register_class<MapItemSingleton>();
+	_map_item_singleton = memnew(MapItemSingleton);
+	Engine::get_singleton()->register_singleton("MapItemSingleton", MapItemSingleton::get_singleton());
 
 	ClassDB::register_class<MenuSingleton>();
 	_menu_singleton = memnew(MenuSingleton);
@@ -116,6 +122,9 @@ void uninitialize_openvic_types(ModuleInitializationLevel p_level) {
 
 	Engine::get_singleton()->unregister_singleton("GameSingleton");
 	memdelete(_game_singleton);
+
+	Engine::get_singleton()->unregister_singleton("MapItemSingleton");
+	memdelete(_map_item_singleton);
 
 	Engine::get_singleton()->unregister_singleton("MenuSingleton");
 	memdelete(_menu_singleton);
--- a/extension/src/openvic-extension/singletons/MapItemSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/MapItemSingleton.cpp
@@ -1,0 +1,237 @@
+#include "MapItemSingleton.hpp"
+#include <string_view>
+
+#include "godot_cpp/variant/packed_int32_array.hpp"
+#include "godot_cpp/variant/packed_vector2_array.hpp"
+#include "godot_cpp/variant/typed_array.hpp"
+#include "godot_cpp/variant/utility_functions.hpp"
+#include "godot_cpp/variant/vector2.hpp"
+#include "openvic-extension/singletons/GameSingleton.hpp"
+#include "openvic-extension/utility/ClassBindings.hpp"
+#include "openvic-extension/utility/Utilities.hpp"
+#include "openvic-simulation/DefinitionManager.hpp"
+#include "openvic-simulation/country/CountryDefinition.hpp"
+#include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/interface/GFXObject.hpp"
+#include "openvic-simulation/map/ProvinceDefinition.hpp"
+#include "openvic-simulation/map/ProvinceInstance.hpp"
+#include "openvic-simulation/map/State.hpp"
+#include "openvic-simulation/types/Vector.hpp"
+
+using namespace godot;
+using namespace OpenVic;
+
+void MapItemSingleton::_bind_methods() {
+	OV_BIND_METHOD(MapItemSingleton::get_billboards);
+	OV_BIND_METHOD(MapItemSingleton::get_province_positions);
+	OV_BIND_METHOD(MapItemSingleton::get_capital_count);
+	OV_BIND_METHOD(MapItemSingleton::get_capital_positions);
+	OV_BIND_METHOD(MapItemSingleton::get_crime_icons);
+	OV_BIND_METHOD(MapItemSingleton::get_rgo_icons);
+	OV_BIND_METHOD(MapItemSingleton::get_national_focus_icons);
+}
+
+MapItemSingleton* MapItemSingleton::get_singleton() {
+	return singleton;
+}
+
+MapItemSingleton::MapItemSingleton() {
+	ERR_FAIL_COND(singleton != nullptr);
+	singleton = this;
+}
+
+MapItemSingleton::~MapItemSingleton() {
+	ERR_FAIL_COND(singleton != this);
+	singleton = nullptr;
+}
+
+// Get the billboard object from the loaded objects
+
+GFX::Billboard const* MapItemSingleton::get_billboard(std::string_view name, bool error_on_fail) const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, nullptr);
+
+	GFX::Billboard const* billboard =
+		game_singleton->get_definition_manager().get_ui_manager().get_cast_object_by_identifier<GFX::Billboard>(name);
+
+	if (error_on_fail) {
+		ERR_FAIL_NULL_V_MSG(billboard, nullptr, vformat("Failed to find billboard \"%s\"", Utilities::std_to_godot_string(name)));
+	}
+
+	return billboard;
+}
+
+// repackage the billboard object into a godot dictionnary for the Billboard manager to work with
+bool MapItemSingleton::add_billboard_dict(std::string_view name, TypedArray<Dictionary>& billboard_dict_array) const {
+
+	static const StringName name_key = "name";
+	static const StringName texture_key = "texture";
+	static const StringName scale_key = "scale";
+	static const StringName noOfFrames_key = "noFrames";
+
+	GFX::Billboard const* billboard = get_billboard(name,false);
+
+	ERR_FAIL_NULL_V_MSG(
+		billboard, false, "Failed to find billboard"
+	);
+
+	Dictionary dict;
+
+	dict[name_key] = Utilities::std_to_godot_string(billboard->get_name());
+	dict[texture_key] = Utilities::std_to_godot_string(billboard->get_texture_file());
+	dict[scale_key] = billboard->get_scale().to_float();
+	dict[noOfFrames_key] = billboard->get_no_of_frames();
+
+	billboard_dict_array.push_back(dict);
+
+	return true;
+}
+
+
+//get an array of all the billboard dictionnaries
+TypedArray<Dictionary> MapItemSingleton::get_billboards() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, {});
+
+	TypedArray<Dictionary> ret;
+
+	for (std::unique_ptr<GFX::Object> const& obj : game_singleton->get_definition_manager().get_ui_manager().get_objects()) {
+		if (obj->is_type<GFX::Billboard>()) {
+			add_billboard_dict(obj->get_name(), ret);
+		}
+	}
+	
+	return ret;
+}
+
+PackedVector2Array MapItemSingleton::get_province_positions() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, PackedVector2Array());
+
+	PackedVector2Array billboard_pos {};
+	
+	for(ProvinceDefinition const& prov : game_singleton->get_definition_manager().get_map_definition().get_province_definitions()){
+		if(prov.is_water()) continue; //billboards dont appear over water, skip
+
+		fvec2_t city_pos = prov.get_city_position();
+		Vector2 pos = Utilities::to_godot_fvec2(city_pos) / game_singleton->get_map_dims();
+		billboard_pos.push_back(pos);
+
+	}
+	
+	return billboard_pos;
+}
+
+//includes non-existent countries, used for setting the billboard buffer size
+int32_t MapItemSingleton::get_capital_count() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, 0);
+
+	return game_singleton->get_definition_manager().get_country_definition_manager().get_country_definition_count();
+}
+
+PackedVector2Array MapItemSingleton::get_capital_positions() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, PackedVector2Array());
+
+	InstanceManager const* instance_manager = game_singleton->get_instance_manager();
+	ERR_FAIL_NULL_V(instance_manager, PackedVector2Array());
+
+	PackedVector2Array billboard_pos {};
+
+	for(CountryInstance const& country : instance_manager->get_country_instance_manager().get_country_instances()){
+		if(!country.exists()) continue; //skip non-existant countries
+
+		fvec2_t city_pos = country.get_capital()->get_province_definition().get_city_position();
+		Vector2 pos = Utilities::to_godot_fvec2(city_pos) / game_singleton->get_map_dims();
+		billboard_pos.push_back(pos);
+
+	}
+
+	return billboard_pos;
+}
+
+PackedByteArray MapItemSingleton::get_crime_icons() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, PackedByteArray());
+
+	InstanceManager const* instance_manager = game_singleton->get_instance_manager();
+	ERR_FAIL_NULL_V(instance_manager, PackedByteArray());
+
+	PackedByteArray icons {};
+
+	for(ProvinceInstance const& prov_inst : instance_manager->get_map_instance().get_province_instances()){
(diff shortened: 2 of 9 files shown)


Nemrav:
Billboards now working with the new mapmodes menu. Ready for review.
Known potential issues:
- Capital billboards appear in front of map mode (ex. RGO) billboards instead of behind, unlike vic2
- mapmode billboards can still sometimes appear in the wrong order (ie. rear ones drawn in front of closer ones). Unsure how to fix this
- Billboard size might be off

Future work:
- Verify Crime billboards only show up on provinces with crime (needs sim update to have provinces w/o crime)
- National Focus billboards once National Focuses are implemented. National Focus icons appear in a state's first province (by index).

Things to verify for review:
- Double check I'm not making an off-by-one mistake in my billboard count, either for the provinces or the country capitals.

extension/src/openvic-extension/register_types.cpp
```diff
@@ -117,6 +124,9 @@ void uninitialize_openvic_types(ModuleInitializationLevel p_level) {
 	Engine::get_singleton()->unregister_singleton("GameSingleton");
 	memdelete(_game_singleton);
 
+		Engine::get_singleton()->unregister_singleton("MapItemSingleton");
```
Spartan322, line 127:
```suggestion
	Engine::get_singleton()->unregister_singleton("MapItemSingleton");
```

Nemrav: good catch

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+	singleton = nullptr;
+}
+
+// Get the billboard object from the loaded objects
+
+GFX::Billboard const* MapItemSingleton::get_billboard(std::string_view name, bool error_on_fail) const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, nullptr);
+	DefinitionManager const* definition_manager = &game_singleton->get_definition_manager();
+	ERR_FAIL_NULL_V(definition_manager, nullptr);
```
Hop311, line 59: No need for this, the definition manager always exists so long as `GameSingleton` exists, so you can just use it directly as `game_singleton->get_definition_manager().get_ui_manager()...`.
The same applies for functions later in this file, you never need to make a `DefinitionManager const*` variable and check it isn't null, that's only for `GameSingleton` and `InstanceManager`

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+	GFX::Billboard const* billboard =
+		definition_manager->get_ui_manager().get_cast_object_by_identifier<GFX::Billboard>(name);
+
+	if (error_on_fail) {
+		//ERR_FAIL_NULL_V_MSG(billboard, nullptr, vformat("Failed to find billboard \"%s\"", Utilities::std_to_godot_string(name)));
+		ERR_FAIL_NULL_V_MSG(billboard, nullptr, "Failed to find billboard inner");
```
Hop311, line 66:
```suggestion
		ERR_FAIL_NULL_V_MSG(billboard, nullptr, vformat("Failed to find billboard \"%s\"", Utilities::std_to_godot_string(name)));
```
Might as well use the error message including the name of the billboard that wasn't found, and remove the other error message.

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+PackedVector2Array MapItemSingleton::get_province_positions(){
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, PackedVector2Array());
+
+	DefinitionManager const* definition_manager = &game_singleton->get_definition_manager();
+	ERR_FAIL_NULL_V(definition_manager, PackedVector2Array());
+
+	PackedVector2Array billboard_pos = PackedVector2Array();
```
Hop311, line 138: Not a big deal, but it isn't necessary to explicitly construct `PackedVector2Array` like this. It's enough to just use `PackedVector2Array billboard_pos {};` here and `ERR_FAIL_NULL_V(game_singleton, {});` above. Same in other fucntions like `get_capital_positions` and `get_crime_icons`

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+	int prov_count = definition_manager->get_map_definition().get_province_definition_count();
```
Hop311, line 140: Unused variable, remove

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+		if (obj->is_type<GFX::Billboard>()) {
+			add_billboard_dict(obj->get_name(), ret);
+		}
+	}
+	
+	return ret;
+}
+
+//dont use this, it includes water provinces
+int MapItemSingleton::get_province_count(){
```
Hop311, line 121: Better to always be explicit with the integer type size, so `int32_t` here and for `get_capital_count`. These should also be changed in the hpp file

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,229 @@
...
+	InstanceManager const* instance_manager = game_singleton->get_instance_manager();
+	ERR_FAIL_NULL_V(instance_manager, PackedInt32Array());
+
+	PackedInt32Array icons = PackedInt32Array();
+
+	for(ProvinceInstance const& prov_inst : instance_manager->get_map_instance().get_province_instances()){
+		ProvinceDefinition const* prov = &prov_inst.get_province_definition();
+		if(prov->is_water()) continue; //billboards dont appear over water, skip
+
+		icons.push_back(prov_inst.get_crime()->get_icon());
```
Hop311, line 199: Currently we give each province a crime for demonstration purposes, but in general we can't assume a province's crime isn't `nullpr`. You'll need to check that here and provide a mechanism for not displaying any icon in that case. The same applies for RGOs, I'm fairly sure it's possible for a province to have no RGO if one isn't given in its history file.

If you can make it so we can selectively hide specific province's icons without needing to re-send the whole list of positions, it would be useful to handle capitals in a similar way: rather than sending a list of all current capitals, we could use a list of all province city positions (this list of positions would be the same for capitals and RGO/crime icons) and then update it with a list of bools indicating whether each province is a capital or not.

Hop311 (commented):
I've reviewed about half of the PR, the main bit left is the billboard GDScript code. However I believe I've identified an important issue/way to improve this system (the ability to selectively hide icons, and so make both capitals and RGO/crime icons use a list of positions for all (land) provinces which are then updated purely based on their status (capital or not & specific RGO/crime icon or no icon) without needing to change the positions after game start.

Hop311:
Also you've added `MapControlPanel.gd` back in this PR after it was removed by Brick (when it was replaced with `Menubar.gd`), please remove it again

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
Nemrav: We decided that having one capital per province wouldn't be efficient. Positions don't need to be re-sent currently with the exception of capitals.

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,39 @@
...
+void fragment() {
+	int image_index = int(COLOR.x);
+	//Color.y == 0 >> invisible, otherwise use the specified frame
+	float visibility = clamp(COLOR.y,0.0,1.0);
+	float frame_index = max(COLOR.y,0.0);
+
+	float uv_x_space = 1.0/float(numframes[image_index]);
+	vec2 uv_out = vec2(uv_x_space * (frame_index + UV.x),UV.y);
+	ALBEDO.rgb = texture( billboards[image_index] ,uv_out).rgb;
+	ALPHA = texture(billboards[image_index] ,uv_out).a * visibility;
```
Hop311, line 38:
```suggestion
	vec4 billboard_col = texture( billboards[image_index] ,uv_out);
	ALBEDO.rgb = billboard_col.rgb;
	ALPHA = billboard_col.a * visibility;
```
Better to look up the pixel once, store it in a variable, and use the values from there, rather than looking it up twice

extension/src/openvic-extension/singletons/MapItemSingleton.hpp
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <godot_cpp/classes/object.hpp>
+#include <godot_cpp/variant/packed_int32_array.hpp>
```
Hop311, line 4: I don't think this is needed anymore, maybe replace it with `packed_byte_array`?

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+//includes non-existent countries, used for setting the billboard buffer size
+int32_t MapItemSingleton::get_capital_count() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, 0);
+
+	InstanceManager const* instance_manager = game_singleton->get_instance_manager();
+	ERR_FAIL_NULL_V(instance_manager, 0);
+
+	return instance_manager->get_country_instance_manager().get_country_instance_count();
```
Hop311, line 145:
```suggestion
	return game_singleton->get_definition_manager().get_country_definition_manager().get_country_definition_count();
```
Use this instead, it gives the same number without needing to worry about instance manager being initialised

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,251 @@
...
+	for (std::unique_ptr<GFX::Object> const& obj : game_singleton->get_definition_manager().get_ui_manager().get_objects()) {
+		if (obj->is_type<GFX::Billboard>()) {
+			add_billboard_dict(obj->get_name(), ret);
+		}
+	}
+	
+	return ret;
+}
+
+//dont use this, it includes water provinces
```
Hop311, line 111: Since this isn't used (and would give a larger value than we need anyway), might as well remove `get_province_count`

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,251 @@
...
+	ERR_FAIL_NULL_V(game_singleton, PackedByteArray());
+
+	InstanceManager const* instance_manager = game_singleton->get_instance_manager();
+	ERR_FAIL_NULL_V(instance_manager, PackedByteArray());
+
+	PackedByteArray icons {};
+
+	for(ProvinceInstance const& prov_inst : instance_manager->get_map_instance().get_province_instances()){
+		ProvinceDefinition const* prov = &prov_inst.get_province_definition();
+		if(prov->is_water()) continue; //billboards dont appear over water, skip
```
Hop311, line 180:
```suggestion
		if (prov_inst.get_province_definition().is_water()) continue; //billboards dont appear over water, skip
```
- `get_province_definition()` returns a `const&`, so no need to turn it into a `const*`
- Since this is only used to check if it's a water province, might as well use it directly rather than creating a variable (this actually makes it take fewer CPU cycles too)
- Same for `get_rgo_icons` and `get_national_focus_icons` below

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+		State const* state = prov_inst.get_state();
+		if(prov->get_identifier() == state->get_capital()->get_identifier()){
```
Hop311, line 242: - Check if the state is null, I don't think it should ever be but better to be safe than sorry
- `state->get_capital()` returns a `ProvinceInstance const*`, rather than comparing the identifier strings (slow) you can just compare the pointers: `&prov_inst == state->get_capital()` (you need to use `&prov_inst` rather than `prov` as you're comparing `ProvinceInstance`s rather than `ProvinceDefinition`s, you could use `prov` if you compare against `&state->get_capital()->get_province_definition()` but that's just slower for no gain)

game/src/Game/GameSession/BillboardManager.gd
```diff
@@ -0,0 +1,191 @@
+extends MultiMeshInstance3D
+
+#given a name: get the index for the texture in the shader
+# this is to reduce the number of magic indeces in the code
+# to get the proper billboard image
+var billboard_names = {}
```
Hop311, line 6: Can you make sure every GDScript `var`, `const` and function argument + return type has an explicit type given

game/src/Game/GameSession/BillboardCamera.gd
```diff
@@ -0,0 +1,18 @@
+extends Camera3D
```
Hop311, line 1: Is `BillboardCamera` actuall used anywhere? I can see there's an instance of it in `GameSession`, but that's not connected/used by anything else from what I can see? Unless it's something to do with the cull mask being `~2` and BillboardManager having layer `2`?

game/src/Game/GameSession/billboards.tres
```diff
@@ -0,0 +1,11 @@
+[gd_resource type="ShaderMaterial" load_steps=2 format=3 uid="uid://dyeip7yhqf1g7"]
```
Hop311, line 1: Is this shader material needed? Would it be possible to use the shader directly without it? We don't have a `tres` file for the map shader, although we may just be creating the material in scripts rather than via the editor like this

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,39 @@
...
+void vertex() {
+	COLOR = INSTANCE_CUSTOM; //send instance_custom info to fragment
+	float size = sizes[int(COLOR.x)];
+	
+	vec3 cam_right_worldspace = vec3(VIEW_MATRIX[0][0],VIEW_MATRIX[1][0],VIEW_MATRIX[2][0]);
+	vec3 cam_up_worldspace = vec3(VIEW_MATRIX[0][1],VIEW_MATRIX[1][1],VIEW_MATRIX[2][1]);
+	vec3 vert_pos_world =
+		cam_right_worldspace * VERTEX.x * size +
+		cam_up_worldspace * VERTEX.y * size;
+	VERTEX = vert_pos_world + vec3(0,height_add_factor*size,0);
```
Hop311, line 26:
```suggestion
	VERTEX =
		cam_right_worldspace * VERTEX.x +
		cam_up_worldspace * VERTEX.y;
	VERTEX.y += height_add_factor;
	VERTEX *= size;
```
I think you can reduce the number of operations needed by rearranging stuff like this (although I haven't tested it fully, so I may be wrong somewhere).
Also is it possible to represent the stuff with `cam_right_worldspace` and `cam_up_worldspace` using matrix multiplcation? Just wondering as I'm sure that will be especially fast for the GPU, even if we're ignoring one of the axes by setting `VERTEX.z` to `0`

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,39 @@
...
+}
+
+void fragment() {
+	int image_index = int(COLOR.x);
+	//Color.y == 0 >> invisible, otherwise use the specified frame
+	float visibility = clamp(COLOR.y,0.0,1.0);
+	float frame_index = max(COLOR.y,0.0);
```
Hop311, line 33: Nice! Very cool shader magic 💯

game/src/Game/GameSession/BillboardManager.gd
```diff
@@ -0,0 +1,191 @@
...
+	else:
+		detailed_map(false)
+
+func _on_map_mode_changed(map_mode : int) -> void:
+	match map_mode:
+		5,6,10: # Infrastructure, Colonial, RGO output
+			set_province_billboards(ProvinceBillboards.RGO)
+		2: #Revolt Risk
+			set_province_billboards(ProvinceBillboards.CRIME)
+		9: #National Focus
```
Hop311, line 181: Maybe create an enum of mapmode indices, even just for use here, to make it cleaner/avoid "magic numbers"

game/src/Game/GameSession/BillboardCamera.gd
Nemrav: To render the billboards on top of everything I found I needed a subviewport. Each subviewport needs its own camera, so I give it a camera, and make it copy the transform of the ordinary camera.
The cull mask is so that only this second camera renders the billboards.

game/src/Game/GameSession/billboard.gdshader
Nemrav: Seems to work.

game/src/Game/GameSession/billboards.tres
Nemrav: Removed the material.

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -0,0 +1,245 @@
+
+
+//billboards, projections, and progress bar
+
```
Spartan322, line 4:
```suggestion
```

game/src/Game/GameSession/BillboardCamera.gd
Nemrav: As per Spartan's comments the subviewport was unnecessary after all and has been removed along with this camera script.

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,45 @@
...
+
+void vertex() {
+	COLOR = INSTANCE_CUSTOM; //send instance_custom info to fragment
+	float size = sizes[int(COLOR.x)];
+	
+	vec3 cam_right_worldspace = vec3(VIEW_MATRIX[0][0],VIEW_MATRIX[1][0],VIEW_MATRIX[2][0]);
+	vec3 cam_up_worldspace = vec3(VIEW_MATRIX[0][1],VIEW_MATRIX[1][1],VIEW_MATRIX[2][1]);
+	VERTEX =
+		cam_right_worldspace * VERTEX.x +
+		cam_up_worldspace * VERTEX.y;
```
Hop311, line 29:
```suggestion
	VERTEX = (vec4(VERTEX, 1.0) * VIEW_MATRIX).xyz;
```
This can be reduced to a matrix multiplication, which on a GPU should be faster than the separate operations it's replacing. Although currently providing `0.0` and `1.0` for the `w` coordinate produces the same results, it's better to go with `1.0` for future-proofing as it's needed for translation transforms to work - `(x, y, z, 1)` represents a position which will be moved by translations, while `(x, y, z, 0)` represents a direction and so is invariant under translations.

Hop311 (commented):
I'm still getting some serious visual bugs with icons jumping/flickering between different textures, e.g. capitals vs yellow teardrop shaped icons, and other stuff like RGO icons appearing on the militancy mapmode

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,45 @@
...
+	VERTEX.y += height_add_factor;
+	VERTEX *= size;
+}
+
+void fragment() {
+	int image_index = int(COLOR.x);
```
Hop311, line 35:
```suggestion
	int image_index = int(COLOR.x + 0.5);
```
Adding `0.5` fixes the z-fighting issue on my Nvidia GPU

Spartan322: that's probably down to the GPU's floating point unit performing truncation differently.

game/src/Game/GameSession/billboard.gdshader
Nemrav: Works.

game/src/Game/GameSession/billboard.gdshader
Nemrav: Works.

Hop311 (dismissed):

game/src/Game/GameSession/billboard.gdshader
```diff
@@ -0,0 +1,41 @@
...
+render_mode unshaded, depth_test_disabled;
+
+
+//vic2 only ever loads a max of 12 BillboardType
+//Of these, only 4 are actually used
+uniform sampler2D billboards[12] : source_color;
+uniform uint numframes[12];
+uniform float sizes[12];
+
+uniform float height_add_factor = 0.3;
```
Hop311, line 14: I don't think this is needed, when I remove it the icons appear closer to their Vic2 positions, whereas with it they're all a bit further to the north

PR #305, opened 2024-11-27
Overhaul style and formatting

Spartan322:
Add pre-commit configuration file
Organize and overhaul .clang-format
Add codespell checks and corrections


extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
```diff
@@ -114,11 +110,11 @@ Error GFXButtonStateTexture::generate_state_image(
...
-	const auto colour_func =
-		button_state == HOVER ? hover_colour :
-		button_state == PRESSED ? pressed_colour :
-		button_state == DISABLED ? disabled_colour :
-		selected_colour;
+	const auto colour_func = button_state == HOVER ? hover_colour
+		: button_state == PRESSED				   ? pressed_colour
+		: button_state == DISABLED
+		? disabled_colour
+		: selected_colour;
```
Hop311, line 117: - Using tab characters to align multiple lines together only works consistently when you're using the same number of characters starting from the same offset, otherwise it breaks when using an interface with a different tab display size (like it's breaking here for me in Github).
- Why are the first two `?`s kept on the same line but the third one is moved to a newline?

extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
```diff
@@ -60,17 +59,14 @@ Ref<GFXButtonStateTexture> GFXButtonStateTexture::make_gfx_button_state_texture(
 	ERR_FAIL_NULL_V(button_state_texture, nullptr);
 	button_state_texture->set_button_state(button_state);
 	if (source_image.is_valid()) {
-		ERR_FAIL_COND_V(
-			button_state_texture->generate_state_image(source_image, region, cornered_tile_border_size) != OK, nullptr
-		);
+		ERR_FAIL_COND_V(button_state_texture->generate_state_image(source_image, region, cornered_tile_border_size) != OK, nullptr);
```
Hop311, line 62: This line has been recombined to now have a column width over 128, while the function call on line 28 wasn't recombined fully (presumably as that would make it over 128 columns wide). Is there any way to make this consistent?

extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
```diff
@@ -141,16 +137,11 @@ StringName const& GFXButtonStateTexture::button_state_to_name(ButtonState button
...
-			return name_disabled;
-		case SELECTED:
-			return name_selected;
-		default:
-			return name_error;
+	case HOVER:	   return name_hover;
+	case PRESSED:  return name_pressed;
+	case DISABLED: return name_disabled;
+	case SELECTED: return name_selected;
+	default:	   return name_error;
```
Hop311, line 144: Again, using tabs for complex alignment produces inconsistent behaviour, using spaces between the `:` and `return` statements would work better.

extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
```diff
@@ -181,8 +172,7 @@ void GFXButtonStateHavingTexture::_clear_button_states() {
 
 GFXButtonStateHavingTexture::GFXButtonStateHavingTexture() : button_image {}, button_state_textures {} {}
 
-Ref<GFXButtonStateTexture> GFXButtonStateHavingTexture::get_button_state_texture(
-	GFXButtonStateTexture::ButtonState button_state
+Ref<GFXButtonStateTexture> GFXButtonStateHavingTexture::get_button_state_texture(GFXButtonStateTexture::ButtonState button_state
 ) {
```
Hop311, line 176: This is breaking just the `) {` onto a new line to avoid going over 128 columns wide, but IMO it looks far better either with no line break or with breaks either side of the `()` around the arguments, as it is now I find it difficult to read at a glance.

extension/src/openvic-extension/classes/GFXMaskedFlagTexture.cpp
```diff
@@ -13,12 +13,10 @@ Error GFXMaskedFlagTexture::_generate_combined_image() {
...
 	 * and so can update it without creating and setting a new image, or not. */
-	bool can_update = button_image.is_valid() && button_image->get_size() == overlay_image->get_size()
-		&& button_image->get_format() == overlay_image->get_format();
+	bool can_update = button_image.is_valid() && button_image->get_size() == overlay_image->get_size() &&
+		button_image->get_format() == overlay_image->get_format();
 	if (!can_update) {
-		button_image = Image::create(
-			overlay_image->get_width(), overlay_image->get_height(), false, overlay_image->get_format()
-		);
+		button_image = Image::create(overlay_image->get_width(), overlay_image->get_height(), false, overlay_image->get_format());
```
Hop311, line 19: This has been recombined to go over the 128 column limit

extension/src/openvic-extension/classes/GFXMaskedFlagTexture.cpp
```diff
@@ -33,23 +31,16 @@ Error GFXMaskedFlagTexture::_generate_combined_image() {
...
 		const Vector2i centre_translation = (mask_image->get_size() - button_image->get_size()) / 2;
 
-		for (
-			Vector2i combined_image_point { 0, 0 };
-			combined_image_point.y < button_image->get_height();
-			++combined_image_point.y
-		) {
-
+		for (Vector2i combined_image_point { 0, 0 }; combined_image_point.y < button_image->get_height();
+			 ++combined_image_point.y) {
```
Hop311, line 35: Here and with the `if` statement below, I'm not a fan of single breaks in the middle of a pair of brackets, I find it much clearer to always have matching breaks after `(` and before `)` if any break is needed between the brackets. I believe therre is a setting that does this for `{}` curly brackets very well, but idk if there's a `()` equivalent.

extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -157,46 +154,29 @@ void GUILabel::_notification(int what) {
...
-	default_colour {},
-	colour_codes { nullptr },
-	currency_texture {},
-	background {},
-	lines {},
-	line_update_queued { false } {}
+	: tooltip_active { false }, gui_text { nullptr }, text {}, substitution_dict {},
+	  horizontal_alignment { HORIZONTAL_ALIGNMENT_LEFT }, max_size {}, border_size {}, adjusted_rect {},
+	  auto_adjust_to_content_size { false }, font {}, font_size { DEFAULT_FONT_SIZE }, default_colour {},
+	  colour_codes { nullptr }, currency_texture {}, background {}, lines {}, line_update_queued { false } {}
```
Hop311, line 179: It's much easier to read long initialisation lists when each argument has its own line, especially when checking if something is missing, or when adding or removing entries.

extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -724,8 +693,7 @@ std::vector<GUILabel::line_t> GUILabel::wrap_lines(std::vector<line_t>& unwrappe
 							if (last_marker_pos != 0 || !current_line->segments.empty()) {
 								if (!new_segment_string.is_empty()) {
 									current_line->segments.emplace_back(string_segment_t {
-										std::move(new_segment_string), string_segment->colour, new_segment_width
-									});
+										std::move(new_segment_string), string_segment->colour, new_segment_width });
```
Hop311, line 696: Ah I thought the `{}` handling would break before the `}` too, I'm fairly sure the old formatting settings had something that did that?

extension/src/openvic-extension/classes/GUIOverlappingElementsBox.cpp
```diff
@@ -32,20 +34,14 @@ Error GUIOverlappingElementsBox::_update_child_positions() {
...
-			break;
-		default:
-			UtilityFunctions::push_error(
-				"Invalid GUIOverlappingElementsBox alignment: ",
-				static_cast<int32_t>(gui_overlapping_elements_box->get_format())
-			);
-			err = FAILED;
+	case left:	 break;
+	case centre: starting_x = (box_width - total_width) / 2; break;
+	case right:	 starting_x = box_width - total_width; break;
```
Hop311, line 39: IMO we should line break before the `break;` (and after the `case ...:` too), I believe there's a formatting setting about having multiple statements on the same line

extension/src/openvic-extension/classes/GUIScrollbar.cpp
```diff
@@ -369,37 +363,43 @@ Error GUIScrollbar::set_gui_scrollbar(GUI::Scrollbar const* new_gui_scrollbar) {
...
-		String const& target, _Element const* element, Ref<GFXSpriteTexture>& texture
-	) -> bool {
-		ERR_FAIL_NULL_V_MSG(element, false, vformat(
-			"Invalid %s element for GUIScrollbar %s - null!", target, gui_scrollbar_name
-		));
+	const auto set_texture =
+		[&gui_scrollbar_name]<typename _Element>(String const& target, _Element const* element, Ref<GFXSpriteTexture>& texture)
+		-> bool {
+		ERR_FAIL_NULL_V_MSG(element, false, vformat("Invalid %s element for GUIScrollbar %s - null!", target, gui_scrollbar_name));
 		const String element_name = Utilities::std_to_godot_string(element->get_name());
```
Hop311, line 370: IMO the lambda function's body should be indented one level further than it's capture/arguments section, and I would prefer not have a line break in the middle of `) -> bool {`

extension/src/openvic-extension/singletons/GameSingleton.cpp
```diff
@@ -257,9 +249,10 @@ Error GameSingleton::_update_colour_image() {
 	Error err = OK;
 
 	InstanceManager const* instance_manager = get_instance_manager();
-	if (instance_manager != nullptr && !get_definition_manager().get_mapmode_manager().generate_mapmode_colours(
-		instance_manager->get_map_instance(), mapmode, colour_data_array.ptrw()
-	)) {
+	if (instance_manager != nullptr &&
+		!get_definition_manager().get_mapmode_manager().generate_mapmode_colours(
+			instance_manager->get_map_instance(), mapmode, colour_data_array.ptrw()
+		)) {
```
Hop311, line 255: The lack of a line break before the `) {` makes the indentation look weird, with the `if` statement's body being the same indentation level as the line above it

extension/src/openvic-extension/singletons/GameSingleton.cpp
```diff
@@ -387,8 +378,8 @@ void GameSingleton::set_viewed_country_by_province_index(int32_t province_index)
 	InstanceManager* instance_manager = get_instance_manager();
 	ERR_FAIL_NULL(instance_manager);
 
-	ProvinceInstance const* province_instance =
-		instance_manager->get_map_instance().get_province_instance_by_index(province_index);
+	ProvinceInstance const* province_instance = instance_manager->get_map_instance().get_province_instance_by_index(province_index
+	);
```
Hop311, line 382: The `);` on its own at the end of the line here is very weird, breaking after the `=` or at both ends of the `()` would look better IMO

Hop311 (commented):
There are a lot of good catches here, particularly spelling mistakes, but in many places clang-format is still very messy and inconsistent IMO, especially with newlines and indentation. If we can find settings that address these issues that would be great, but in its current state I find it makes the code harder to read overall and I wouldn't want it applied everywhere automatically.

extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
Spartan322: Because clang-format has literally no idea what is going on after the first two ternary operators.

extension/src/openvic-extension/classes/GFXButtonStateTexture.cpp
Spartan322: No because clang-format has no semantic information, its decisions are completely based on independent line decisions.

extension/src/openvic-extension/classes/GUIScrollbar.cpp
Spartan322: There is literally no formatter capable of that because that lambda definition doesn't reside in an indentation.

PR #306, opened 2024-11-27
Document state merging

### LINKED ISSUE #268: Document how merging states works
### Problem

In Victoria 2 a state can be split between countries. An example of this is Osthannover being split between Bremen, Hamburg and Hannover. All three countries can build factories and a port in their part of the state. When the state is united, there may only be 1 port and max 8 factories. How does Victoria 2 decide which port and which factories to keep?

OpenVic needs to faithfully recreate this behaviour. The inner workings are currently unknown and need to be discovered and documented.

### Solution

Experiment with uniting states in Victoria 2 to figure out the following:

- Does building level matter?
- What happens to duplicate factories?
- Does it matter whether a factory is closed or open?
- Does the subsidies checkbox keep its value?

If you find other things that play a role here, document them as well.
When using other people's findings, from project alice, reddit or paradox forums for example, always verify their claims in Victoria 2.

Document your findings in [OpenVic/docs/simulation/](https://github.com/OpenVicProject/OpenVic/tree/master/docs/simulation).

For questions, just ask me on Discord.

zaaarf:
As requested in #268.

I recommend someone else double checks this, because there are a number of things to account for and I may have missed something.

### DIFF
--- a/docs/simulation/state-merging.md
+++ b/docs/simulation/state-merging.md
@@ -1,0 +1,14 @@
+# State merging
+In the game, it may happen that two different countries own provinces that are part of a same state. Suppose a situation where a country acquires the remainder of a state it already owned. What happens to the buildings?
+
+## Ports
+The port rules are straightforward. The highest-level port is retained, and the rest are deleted.
+
+## Factories
+Factory logic is quite complicated. If there are two factories of the same type in the states to be merged, the following rules are applied in order:
+
+- If the factory in the state that was already owned is subsidized, it takes priority, regardless of level.
+- The factory with the highest level takes precedence.
+- The factory whom the country already owned is chosen, if it comes to a stalemate.
+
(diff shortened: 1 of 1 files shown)


docs/simulation/state-merging.md
```diff
@@ -0,0 +1,14 @@
+# State merging
+In the game, it may happen that two different countries own provinces that are part of a same state. Suppose a situation where a country acquires the remainder of a state it already owned. What happens to the buildings?
+
+## Ports
+The port rules are straightforward. The highest-level port is retained, and the rest are deleted.
+
+## Factories
+Factory logic is quite complicated. If there are two factories of the same type in the states to be merged, the following rules are applied in order:
+
+- If the factory in the state that was already owned is subsidized, it takes priority, regardless of level.
```
wvpm, line 10: When I tried, the existing factory always stays regardless of subsidies, hiring priority or level.

PR #320, opened 2024-12-08
Add shortcut GUI support

Spartan322:
Depends on OpenVicProject/OpenVic-Simulation#244

Add pause menu shortcut
Add time speed up, time speed down, and time pause shortcuts

### DIFF
--- a/extension/src/openvic-extension/utility/UITools.cpp
+++ b/extension/src/openvic-extension/utility/UITools.cpp
@@ -1,7 +1,9 @@
 #include "UITools.hpp"
+#include <cctype>
 
 #include <godot_cpp/classes/color_rect.hpp>
 #include <godot_cpp/classes/line_edit.hpp>
+#include <godot_cpp/classes/os.hpp>
 #include <godot_cpp/classes/panel.hpp>
 #include <godot_cpp/classes/style_box_empty.hpp>
 #include <godot_cpp/classes/style_box_texture.hpp>
@@ -22,6 +24,10 @@
 #include "openvic-extension/singletons/AssetManager.hpp"
 #include "openvic-extension/singletons/GameSingleton.hpp"
 #include "openvic-extension/utility/Utilities.hpp"
+#include "godot_cpp/classes/global_constants.hpp"
+#include "godot_cpp/classes/input_event_key.hpp"
+#include "godot_cpp/classes/shortcut.hpp"
+#include "godot_cpp/core/error_macros.hpp"
 
 using namespace godot;
 using namespace OpenVic;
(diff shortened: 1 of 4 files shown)


extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -217,9 +277,15 @@ static bool generate_icon(generate_gui_args_t&& args) {
...
-	// TODO - shortcut, clicksound, rotation (?)
+	// TODO - clicksound, rotation (?)
 	const String button_name = Utilities::std_to_godot_string(button.get_name());
+	const String shortcut_key_name = Utilities::std_to_godot_string(button.get_shortcut());
+	Array event_array = get_events_from_shortcut_key(shortcut_key_name);
 
+	ERR_FAIL_COND_V_MSG(
+		shortcut_key_name.length() != 0 && event_array.size() == 0, false,
+		vformat("Unknown shortcut key '%s' for GUI button %s", shortcut_key_name, button_name)
+	);
```
Hop311, line 288: Better to print an error message and set `bool ret` to false if the shortcut isn't generated properly, rather than completely fail to generate the button. We do completely fail below in various cases where we can't generate the sprite, which at the time was all the content the button was made up of, but in retrospect it's also better not to completely fail element generation due to a missing texture.

IMO we should modify this function to allow the `godot::Button` to generate even if there's an issue with the shortcut or texture, and the same for `generate_checkbox`, but these changes don't necessarily have to happen in this PR.

Hop311 (approved):
WASD now affect both mapmodes and cardinal movement (but only when pressed and held, unless you also hold shift), I'm leaning towards getting rid of WASD cardinal movement as it's cumbersome to press and hold, and the changing mapmodes while using it are annoying too. Better to stick with the arrows keys IMO, same as Vic2. Thinking some more, another possibility is to keep shift+WASD and only remove the regular WASD, so you can choose when to use it rather than being forced to mix mapmodes & movement.

extension/src/openvic-extension/utility/UITools.cpp
Spartan322: Yeah we should fix that later as something don't altogether.

extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -60,6 +66,60 @@ GUI::Position const* UITools::get_gui_position(String const& gui_scene, String c
...
+		} else if (key.nocasecmp_to("PAGE_UP") == 0) {
+			key_value = Key::KEY_PAGEUP;
+		} else if (key.nocasecmp_to("PAGE_DOWN") == 0) {
+			key_value = Key::KEY_PAGEDOWN;
+		} else if (key == "+") {
+			// on most keyboards + and = are on the same key, Godot does not see them as different
+			key_value = Key::KEY_EQUAL;
+		} else if (key == "-") {
+			key_value = Key::KEY_MINUS;
+		} else if (key == ">") {
```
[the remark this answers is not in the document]
Spartan322: It didn't appear common for vanilla or mods to use that, I had to limit the scope of doing such cause there isn't really another way to check for whether a specific key event would produce a unicode value with shortcuts that works. (InputEventKey's `unicode` property is not ever evaluated)

Spartan322:
I'll get rid of the WASD, will keep the shift and arrow keys.

Hop311 (approved):

PR #322, opened 2024-12-08
Fix province icon scaling

Hop311:
- Province icons (RGOs, crimes, national foci) are now scaled by 0.75 vertically, while capitals are still 1:1.
- Some tweaks to billboard management:
  - we only set up billboards we actually use, the others are ignored.
  - `StringName`s are used over `String`s where possible.
  - `ProvinceBillboards` enum values are used where possible rather than hardcoded strings.

### DIFF
--- a/game/src/Game/GameSession/BillboardManager.gd
+++ b/game/src/Game/GameSession/BillboardManager.gd
@@ -1,32 +1,48 @@
 extends MultiMeshInstance3D
 
-#given a name: get the index for the texture in the shader
-# this is to reduce the number of magic indeces in the code
-# to get the proper billboard image
-var billboard_names : Dictionary = {}
-
 @export var _map_view : MapView
 
-const SCALE_FACTOR : float = 1.0/96.0
+const SCALE_FACTOR : float = 1.0 / 96.0
 
-enum ProvinceBillboards { NONE, INVISIBLE, RGO, CRIME, NATIONAL_FOCUS }
+enum BillboardType { NONE, RGO, CRIME, NATIONAL_FOCUS, CAPITAL }
+const BILLBOARD_NAMES : Dictionary = {
+	BillboardType.RGO: &"tradegoods",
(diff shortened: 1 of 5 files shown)


game/src/Game/GameSession/BillboardManager.gd
```diff
@@ -1,23 +1,31 @@
...
 @export var _map_view : MapView
 
-const SCALE_FACTOR : float = 1.0/96.0
-
-enum ProvinceBillboards { NONE, INVISIBLE, RGO, CRIME, NATIONAL_FOCUS }
+const SCALE_FACTOR : float = 1.0 / 96.0
+const ICON_DIMS : Vector2 = Vector2(1.0, 0.75)
+const CAPITAL_DIMS : Vector2 = Vector2(1.0, 1.0)
+
+enum ProvinceBillboards { NONE, INVISIBLE, RGO, CRIME, NATIONAL_FOCUS, CAPITAL }
```
Nemrav, line 9: The original intent was to represent the current state of the province billboards. Now that it also includes CAPITAL, it should be renamed something more generic like `BillboardState`. It doesn't have to be that name, just something better descriptive of what the enum is for.

Nemrav (changes requested):
Just a couple small tweaks, nice work.

Hop311:
I think something else might be stretching our icons vertically beyond the dims we set, as Vic2 definitely scales RGO icons by exactly 0.8 vertically but in OpenVic the same value makes them look noticably taller. Similarly, Vic2 doesn't seem to stretch crime icons vertically, but ours appear slightly taller than they should even at 1:1 scale.

Also some icons aren't appearing in the correct position, e.g. the entire south coast of the UK (Plymouth, Salisbury, Southampton, Brighton and Canterbury) have their icons in different positions in Vic2 vs OpenVic.

Nemrav (approved):
Looks Good, Nice work.

PR #326, opened 2024-12-14
Add player customizable button hotkeys

Spartan322:
Fix potential button generation fail for invalid hotkeys


extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -66,7 +71,7 @@ GUI::Position const* UITools::get_gui_position(String const& gui_scene, String c
 	return position;
 }
 
-static Array get_events_from_shortcut_key(String const& key) {
+Array get_events_from_shortcut_key(String const& key) {
```
Hop311, line 74: Why not make `get_events_from_shortcut_key` and `try_create_shortcut_action_for_button` static functions, local to `UITools.cpp`?

extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -123,6 +128,51 @@ static Array get_events_from_shortcut_key(String const& key) {
...
 
+Error try_create_shortcut_action_for_button(GUIButton* gui_button, String const& shortcut_key_name, String const& shortcut_hotkey_name = "") {
+	Array event_array = get_events_from_shortcut_key(shortcut_key_name);
+
+	ERR_FAIL_COND_V_MSG(
+		shortcut_key_name.length() != 0 && event_array.size() == 0, ERR_INVALID_PARAMETER,
+		vformat("Unknown shortcut key '%s' for GUI button %s", shortcut_key_name, gui_button->get_name())
+	);
+
+	if (shortcut_key_name.length() == 0) {
```
Hop311, line 139:
```suggestion
	if (shortcut_key_name.is_empty()) {
```
Could use `is_empty` here and at the beginning of `get_events_from_shortcut_key`, and `!shortcut_key_name.is_empty()` in the error check above

extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -123,6 +128,51 @@ static Array get_events_from_shortcut_key(String const& key) {
...
+		return OK;
+	}
+
+	InputMap* const im = InputMap::get_singleton();
+	String action_name;
+	if (shortcut_hotkey_name.is_empty()) {
+		action_name = //
+			vformat("button_%s_hotkey", gui_button->get_name().to_lower().replace("button", "").replace("hotkey", ""))
+				.replace("__", "_");
```
Hop311, line 148: What about a leftover `_` at either end of the name? Currently `close_button` -> `close_`, rather than `close`. Maybe call `replace("__", "_")` after `vformat`, even as well as inside `vformat` (to avoid any risk of a triple `_` that can't be fixed with one `replace`, unless `replace` can actualy handle that case)

Hop311 (commented):
- On Windows the initially loaded keybindings are fine but anything I change them to doesn't work and the original biinding keeps working instead. Also the original binding has "(Unicode)" after it in the options menu controls UI which disappears after changing the binding, not even coming back when you type in the original binding manually.
- This doesn't allow setting shortcuts on anything without a shortcut already set in the defines, even `shortcut = ""` won't generate an action that can be customised

extension/src/openvic-extension/utility/UITools.cpp
Spartan322: Because there is no reason for the function to be accessible to any other translation unit.

extension/src/openvic-extension/utility/UITools.cpp
Spartan322: > Maybe call replace("__", "_") after vformat

It is.

Spartan322:
>  Also the original binding has "(Unicode)" after it in the options menu controls UI which disappears after changing the binding

That is irrelevant, that's just a product of what is required to try and somewhat mimic what Victoria 2 tries to do.

> This doesn't allow setting shortcuts on anything without a shortcut already set in the defines, even shortcut = "" won't generate an action that can be customised

Its not intended to be, this is not intended to add hotkeys to things that don't have them, this only would allow customizing things that already have hotkeys.

> We should start thinking about how to display shortcuts in tooltips, depending on how complicated it is to generate a string representation of a key binding we should either cache or just always create a shortcut section for the tooltip and add it automatically any time the tooltip is updated. We'd need to detect when to update the shortcut text too if it's cached, and maybe refresh the tooltip when bindings change. Even if they're otherwise empty, tooltips should generate with the shortcut displayed at the minimum.

Why would we need to cache it? Godot makes this trivial to do already, it already tries to do it automatically by displaying the first valid event key in a tooltip which is why I call `set_shortcut_in_tooltip(false)` as its obnoxious with our preexisting tooltip, you can just get a textual representation of the InputEvent.

extension/src/openvic-extension/utility/UITools.cpp
Hop311: Yeah that's what I mean, make them have the `static` prefix in `UITools.cpp` so they definitely stay just in this translation unit

Hop311 (approved):
Shortcuts are working including new ones like manually adding `shortcut = "F"` to `button_goto` in `menubar.gui` to open the search panel with shift+F, although I can't see the new actions in the Options/Control menu at all now 😕

extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -123,6 +128,69 @@ static Array get_events_from_shortcut_key(String const& key) {
...
+	Array event_array = get_events_from_shortcut_key(shortcut_key_name);
+
+	ERR_FAIL_COND_V_MSG(
+		!shortcut_key_name.is_empty() && event_array.is_empty(), ERR_INVALID_PARAMETER,
+		vformat("Unknown shortcut key '%s' for GUI button %s", shortcut_key_name, gui_button->get_name())
+	);
+
+	if (shortcut_key_name.is_empty()) {
+		return OK;
+	}
```
Hop311, line 143: Not a big deal, but this could go at the very beginning of the function and skip the first two steps.

extension/src/openvic-extension/utility/UITools.cpp
```diff
@@ -123,6 +128,69 @@ static Array get_events_from_shortcut_key(String const& key) {
...
+		TypedArray<InputEvent> events = im->action_get_events(action_name);
+		bool skip_warning = events.size() != event_array.size();
+		if (!skip_warning) {
+			for (std::size_t index = 0; index < events.size(); index++) {
+				if (!event_array.has(events[index])) {
+					skip_warning = false;
+					break;
+				}
+			}
+		}
```
Hop311, line 169: This block only triggers if `skip_warning` is false, and all it can do is `skip_warning = false`, so either it's unnecessary or there's a mistake?

game/addons/keychain/Keychain.gd
```diff
@@ -22,9 +24,12 @@ var ignore_ui_actions := true
 ## and the fourth for [InputEventJoypadMotion]s.
 var changeable_types: PackedByteArray = [true, true, true, true]
 ## The file path of the [code]config_file[/code].
-var config_path := "user://cache.ini"
+var config_path := "user://config.ini"
 ## Used to store the settings to the filesystem.
 var config_file: ConfigFile
+## Used to check if unused binding check should be ignored for action
+var keep_binding_check : Callable = func(action_name : StringName) -> bool:
+	return action_name.begins_with("button_") and action_name.ends_with("_hotkey")
```
Hop311, line 32: Why is this a `Callable` rather than a function (even a `static` function)?

Hop311 (dismissed):
We get a bunch of warnings from UI elements with the same name but different values, not sure if these are fine to ignore or if we want to be able to handle them separately (e.g. create a `2`-suffixed version if a regular action already exists)
```
WARNING: 'button_rallypoint_checkbox_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_merge_checkbox_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_checkbox_naval_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_merge_checkbox_naval_hotkey' already found in InputMap with different values, reusing hotkey     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_checkbox_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_merge_checkbox_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_checkbox_naval_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_rallypoint_merge_checkbox_naval_hotkey' already found in InputMap with different values, reusing hotkey     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
Mapmode set to "Terrain Mapmode" (index: 0, identifier: mapmode_terrain)
WARNING: 'button_close_hotkey' already found in InputMap with different values, reusing hotkey
     at: try_create_shortcut_action_for_button (extension\src\openvic-extension\utility\UITools.cpp:172)
```

game/addons/keychain/Keychain.gd
Spartan322: Because I'm trying to consider a way to remove reliance on hardcoding the button hotkey check from within Keychain, I want to minimize modifications to it.

Hop311 (approved):

PR #335, opened 2024-12-21
Add map projections

Nemrav:
Adds support for the unit selection and valid/invalid move markers (which in vic2 are ProjectionType).
- To see the selection markers: shift+left-click to add a marker at a unit position. left-click to clear the projections.
- Right click on provinces to add valid/invalid movement markers.
A main limitation is that the selection markers do not follow units, but are placed wherever a province is shift+clicked. This won't be fixeable until we have proper unit selection

### DIFF
--- a/extension/src/openvic-extension/singletons/MapItemSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/MapItemSingleton.cpp
@@ -1,17 +1,18 @@
 #include "MapItemSingleton.hpp"
 #include <string_view>
 
+#include "godot_cpp/core/error_macros.hpp"
 #include "godot_cpp/variant/packed_int32_array.hpp"
 #include "godot_cpp/variant/packed_vector2_array.hpp"
 #include "godot_cpp/variant/typed_array.hpp"
-#include "godot_cpp/variant/utility_functions.hpp"
 #include "godot_cpp/variant/vector2.hpp"
 #include "openvic-extension/singletons/GameSingleton.hpp"
 #include "openvic-extension/utility/ClassBindings.hpp"
 #include "openvic-extension/utility/Utilities.hpp"
-#include "openvic-simulation/DefinitionManager.hpp"
 #include "openvic-simulation/country/CountryDefinition.hpp"
 #include "openvic-simulation/country/CountryInstance.hpp"
+#include "openvic-simulation/DefinitionManager.hpp"
+#include "openvic-simulation/economy/BuildingType.hpp"
 #include "openvic-simulation/interface/GFXObject.hpp"
 #include "openvic-simulation/map/ProvinceDefinition.hpp"
 #include "openvic-simulation/map/ProvinceInstance.hpp"
@@ -20,6 +21,7 @@
 
 using namespace godot;
 using namespace OpenVic;
+using namespace OpenVic::Utilities::literals;
 
 void MapItemSingleton::_bind_methods() {
 	OV_BIND_METHOD(MapItemSingleton::get_billboards);
@@ -29,6 +31,11 @@ void MapItemSingleton::_bind_methods() {
 	OV_BIND_METHOD(MapItemSingleton::get_crime_icons);
 	OV_BIND_METHOD(MapItemSingleton::get_rgo_icons);
 	OV_BIND_METHOD(MapItemSingleton::get_national_focus_icons);
+	OV_BIND_METHOD(MapItemSingleton::get_projections);
+	OV_BIND_METHOD(MapItemSingleton::get_unit_position_by_province_index,{"index"});
+	OV_BIND_METHOD(MapItemSingleton::get_port_position_by_province_index,{"index"});
+	OV_BIND_METHOD(MapItemSingleton::get_clicked_port_province_index, {"position"});
+	
 }
 
 MapItemSingleton* MapItemSingleton::get_singleton() {
@@ -45,57 +52,72 @@ MapItemSingleton::~MapItemSingleton() {
 	singleton = nullptr;
 }
 
-// Get the billboard object from the loaded objects
-GFX::Billboard const* MapItemSingleton::get_billboard(std::string_view name, bool error_on_fail) const {
-	GameSingleton const* game_singleton = GameSingleton::get_singleton();
-	ERR_FAIL_NULL_V(game_singleton, nullptr);
-
-	GFX::Billboard const* billboard =
-		game_singleton->get_definition_manager().get_ui_manager().get_cast_object_by_identifier<GFX::Billboard>(name);
-
-	if (error_on_fail) {
-		ERR_FAIL_NULL_V_MSG(
-			billboard, nullptr, vformat("Failed to find billboard \"%s\"", Utilities::std_to_godot_string(name))
-		);
-	}
-
-	return billboard;
-}
-
 // repackage the billboard object into a godot dictionary for the Billboard manager to work with
-bool MapItemSingleton::add_billboard_dict(std::string_view name, TypedArray<Dictionary>& billboard_dict_array) const {
+void MapItemSingleton::add_billboard_dict(GFX::Billboard const& billboard, TypedArray<Dictionary>& billboard_dict_array) const {
 
 	static const StringName name_key = "name";
 	static const StringName texture_key = "texture";
 	static const StringName scale_key = "scale";
 	static const StringName noOfFrames_key = "noFrames";
 
-	GFX::Billboard const* billboard = get_billboard(name, false);
-
-	ERR_FAIL_NULL_V_MSG(billboard, false, vformat("Failed to find billboard \"%s\"", Utilities::std_to_godot_string(name)));
-
 	Dictionary dict;
 
-	dict[name_key] = Utilities::std_to_godot_string(billboard->get_name());
-	dict[texture_key] = Utilities::std_to_godot_string(billboard->get_texture_file());
-	dict[scale_key] = billboard->get_scale().to_float();
-	dict[noOfFrames_key] = billboard->get_no_of_frames();
+	dict[name_key] = Utilities::std_to_godot_string(billboard.get_name());
+	dict[texture_key] = Utilities::std_to_godot_string(billboard.get_texture_file());
+	dict[scale_key] = billboard.get_scale().to_float();
+	dict[noOfFrames_key] = billboard.get_no_of_frames();
 
 	billboard_dict_array.push_back(dict);
-
-	return true;
 }
 
 //get an array of all the billboard dictionaries
 TypedArray<Dictionary> MapItemSingleton::get_billboards() const {
 	GameSingleton const* game_singleton = GameSingleton::get_singleton();
-	ERR_FAIL_NULL_V(game_singleton, {});
+
 
 	TypedArray<Dictionary> ret;
 
 	for (std::unique_ptr<GFX::Object> const& obj : game_singleton->get_definition_manager().get_ui_manager().get_objects()) {
-		if (obj->is_type<GFX::Billboard>()) {
-			add_billboard_dict(obj->get_name(), ret);
+		GFX::Billboard const* billboard = obj->cast_to<GFX::Billboard>();
+		if (billboard != nullptr) {
+			add_billboard_dict(*billboard, ret);
+		}
+	}
+
+	return ret;
+}
+
+void MapItemSingleton::add_projection_dict(GFX::Projection const& projection, TypedArray<Dictionary>& projection_dict_array) const {
+	static const StringName name_key = "name";
+	static const StringName texture_key = "texture";
+	static const StringName size_key = "size";
+	static const StringName spin_key = "spin";
+	static const StringName expanding_key = "expanding";
+	static const StringName duration_key = "duration";
+	static const StringName additative_key = "additative";
+
+	Dictionary dict;
+
+	dict[name_key] = Utilities::std_to_godot_string(projection.get_name());
+	dict[texture_key] = Utilities::std_to_godot_string(projection.get_texture_file());
+	dict[size_key] = projection.get_size().to_float();
+	dict[spin_key] = projection.get_spin().to_float();
+	dict[expanding_key] = projection.get_expanding().to_float();
+	dict[duration_key] = projection.get_duration().to_float();
+	dict[additative_key] = projection.get_additative();
+
+	projection_dict_array.push_back(dict);
+}
+
+TypedArray<Dictionary> MapItemSingleton::get_projections() const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+
+	TypedArray<Dictionary> ret;
+
+	for (std::unique_ptr<GFX::Object> const& obj : game_singleton->get_definition_manager().get_ui_manager().get_objects()) {
+		GFX::Projection const* projection = obj->cast_to<GFX::Projection>();
+		if (projection != nullptr) {
+			add_projection_dict(*projection, ret);
 		}
 	}
 
@@ -104,8 +126,6 @@ TypedArray<Dictionary> MapItemSingleton::get_billboards() const {
 
 PackedVector2Array MapItemSingleton::get_province_positions() const {
 	GameSingleton const* game_singleton = GameSingleton::get_singleton();
-	ERR_FAIL_NULL_V(game_singleton, {});
-
 	MapDefinition const& map_definition = game_singleton->get_definition_manager().get_map_definition();
 
 	PackedVector2Array billboard_pos {};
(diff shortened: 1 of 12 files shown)


game/src/Game/GameSession/MarkerManager.gd
```diff
@@ -0,0 +1,122 @@
...
+
+	material.set_shader_parameter(&"projections", textures)
+	material.set_shader_parameter(&"sizes", sizes)
+	material.set_shader_parameter(&"spin", spins)
+	material.set_shader_parameter(&"expanding", expansions)
+	material.set_shader_parameter(&"additative", transparency_mode)
+	material.set_shader_parameter(&"duration",durations)
+	material.set_shader_parameter(&"time", times)
+	
+	moveMarkers.multimesh.mesh.surface_set_material(0, material)
```
Hop311, line 106: Out of interest, is this necessary since `material` came from `moveMarkers.multimesh.mesh.surface_get_material(0)`?

game/src/Game/GameSession/MarkerManager.gd
```diff
@@ -0,0 +1,122 @@
...
+	for i : int in times.size():
+		times[i] += delta
+	
+	material.set_shader_parameter(&"time", times)
+	moveMarkers.set_time(times)
+	selectionMarkers.set_time(times[Type_to_Index[ProjectionType.SELECTED]])
+	
+	for i : int in times.size():
+		if times[i] > loop_times[i]:
+			times[i] = 0
```
Hop311, line 122:
```suggestion
		if times[i] >= loop_times[i]:
			times[i] -= loop_times[i]
```
- Might as well reset when equalling `loop_times[i]` rather than waiting to be explicitly above it
- Better to subtract `loop_times[i]` so we're not losing/jumping back by `times[i] - loop_times[i]`

game/src/Game/GameSession/Projection.gdshader
```diff
@@ -0,0 +1,45 @@
+shader_type spatial;
+render_mode unshaded;
+
+uniform sampler2D projections[3] : source_color;
+uniform float sizes[3];
+uniform float spin[3];
+uniform float expanding[3];
+uniform float duration[3];
+uniform bool additative[3]; //if true, black becomes a transparency colour
+uniform float time[3]; //= 0.0;
```
Hop311, line 10:
```suggestion
uniform float time[3];
```

game/src/Game/GameSession/Projection.gdshader
```diff
@@ -0,0 +1,45 @@
...
+	COLOR = INSTANCE_CUSTOM;
+	uint type = uint(COLOR.x + 0.5);
+	float start_time = COLOR.y;
+
+	float rot = time[type]*spin[type];
+	mat3 rotation_matrix = mat3(
+		vec3(cos(-rot), 0.0, sin(-rot)),
+		vec3(0.0, 1.0, 0.0),
+		vec3(-sin(-rot), 0.0, cos(-rot))
+	);
```
Hop311, line 22:
```suggestion
	float cos_rot = cos(rot);
	float sin_rot = sin(rot);
	mat3 rotation_matrix = mat3(
		vec3(cos_rot, 0.0, -sin_rot),
		vec3(0.0, 1.0, 0.0),
		vec3(sin_rot, 0.0, cos_rot)
	);
```
- `sin` is an odd function so `sin(-x) = -sin(x)`, and `cos` is an even function so `cos(-x) = cos(x)`
- Might as well only calculate `sin(rot)` and `cos(rot)` once

game/src/Game/GameSession/Projection.gdshader
```diff
@@ -0,0 +1,45 @@
...
+	VERTEX.xyz *= rotation_matrix * 
+		clamp(
+			expanding[type] * (time[type]-start_time),
+			0.0,
+			sizes[type]
+		);
```
Hop311, line 28: Currently this multiplies a matrix by a constant, which is 9 multiplications, 4 of which are with 0. You could reduce it to 3 by just multiplying `VERTEX.xyz` before or after it's been multiplied by rotation matrix, or you could only do it twice if you:
- calculate `clamp(...)` and store it in a float variable
- multiply `cos_rot` and `sin_rot` by that variable when calculating them
- replace the `1.0` in the matrix with the `clamp(...)` variable

You now have `rotation_matrix` multiplid by `clamp(...)` with only 2 multiplications

game/src/Game/GameSession/Projection.gdshader
```diff
@@ -0,0 +1,45 @@
...
+	ALBEDO.rgb = sample.rgb;
+
+	float start_time = COLOR.y;
+	//duration == 0 -> 1, duration == 1 -> 0
+	float unlimited = float(duration[type] == 0.0);
+	float is_finished = unlimited + ceil(clamp(duration[type] - time[type] + start_time,0.0,1.0));
+
+	//if additative, then black = transparent, otherwise use alpha
+	float add_tr = float(additative[type]);
+	ALPHA = is_finished*((1.0 - add_tr)*sample.a + add_tr*(sample.r+sample.g+sample.b));
```
Hop311, line 44: Rather than manually doing `(1.0 - add_tr) * x + add_tr * y`, you can us `mix` to do the interpolation (always better to use built-in functions as they may use specific GPU hardware)

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+
+var time : float = 0.0
+var loop_time : float = 0.0
+const MIN_INSTANCE_COUNT : int = 32
+# above this size, and clearing the selection will reset things to the min
+const MAX_INSTANCE_COUNT : int = 128 
+#When the number of selections exceeds the current instance count, grow the instance buffer size
+# by this much. Larger the number, the less often the buffer is resized, but there is a less fine grained number of instances.
+const INSTANCE_COUNT_GROW_AMOUNT : int = 8 
+const HEIGHT_ADD_FACTOR : Vector3 = Vector3(0,0.002,0)
```
Hop311, line 14: You could use `ProjectionManager.HEIGHT_ADD_FACTOR` rather than redefining it here

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+			))
+			ids[i] = unit_id
+			slot_found = true
+			break
+			
+	if !slot_found:
+		_grow_buffer()
+		
+		var slot_index : int = ids.size()
+		ids.resize(slot_index-1 + INSTANCE_COUNT_GROW_AMOUNT)
```
Hop311, line 61: Why the `-1`? This leaves `ids` 1 smaller than the multimesh buffer
Also should we check we're not growing beyond `MAX_INSTANCE_COUNT` here?

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+		
+		multimesh.set_instance_transform(
+			slot_index,Transform3D(Basis(), unit_position + HEIGHT_ADD_FACTOR)
+		)
+		multimesh.set_instance_custom_data(slot_index,Color(
+			select_index,time,2.0,0.0
```
Hop311, line 67: I notice the z/b value is `2.0` here and `1.0` above, any reason behind this?

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+		))
+		ids[slot_index] = unit_id
+		for j : int in ids.size() - slot_index:
+			ids[slot_index + j] = -1
+	multimesh.visible_instance_count += 1
+
+func update_selection_marker(unit_id : int, unit_position : Vector3) -> void:
+	var index : int = ids.find(unit_id)
```
Hop311, line 75: Should we check this succeeds (so `index != -1`)?

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+	multimesh.set_instance_transform(
+		index,Transform3D(Basis(), unit_position + HEIGHT_ADD_FACTOR)
+	)
+
+#TODO: Perhaps there's a more efficient method...
+func add_selection_markers(unit_ids : PackedInt32Array, unit_positions : PackedVector3Array) -> void:
+	assert(unit_ids.size() == unit_positions.size())
```
Hop311, line 82: I guess we could use a `PackedVector4Array`, although that might break for arbitrarily large `unit_ids[index]` values

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+	multimesh.visible_instance_count += unit_positions.size()
```
Hop311, line 83: `add_selection_marker` already increases the `multimesh.visible_instance_count` by 1 every time it's called, so we end up increasing it by `2 * unit_positions.size()` here. If it's more efficient to do one big increase rather than many small ones, you could add an argument to `add_selection_marker` which controls whether it increases `multimesh.visible_instance_count` and disable it here

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+		setup()
+		return
+	ids.fill(-1)
+	multimesh.visible_instance_count = 0
+
+
+# we can control the multimesh buffer directly
+# 12 floats for transform, 4 floats for colour, 4 floats for custom data
+# colour and custom data are optional
+const TR_CUSTOM_SIZE : int = 16
```
Hop311, line 104: Do we not use the custom data floats? 16 = 12 for transform + 4 for colour, if we had custom data it would be 20?

game/src/Game/GameSession/ValidMoveMarkers.gd
```diff
@@ -0,0 +1,101 @@
...
+
+var projection_type_to_index : Dictionary
+
+var time_legal : float = 0.0
+var time_illegal : float = 0.0
+var loop_time_legal : float = 0.0
+var loop_time_illegal : float = 0.0
+
+const INSTANCE_COUNT : int = 64
+const HEIGHT_ADD_FACTOR : Vector3 = Vector3(0,0.002,0)
```
Hop311, line 14: Again, you could use `ProjectionManager.HEIGHT_ADD_FACTOR` instead of redefining it

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
+class_name SelectionMarkers
+extends MultiMeshInstance3D
+
+@export var manager : ProjectionManager
+
+var time : float = 0.0
+var loop_time : float = 0.0
+const MIN_INSTANCE_COUNT : int = 32
+# above this size, and clearing the selection will reset things to the min
+const MAX_INSTANCE_COUNT : int = 128
```
Hop311, line 10: We should check if the base game has a limit by selecting as many units as possible at the same time. Worst case we could put the limit above the number of provinces in a normal map

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,119 @@
...
+	multimesh.visible_instance_count = 0
+
+# for markers with unlimited duration, time will rollover eventually. so we need a way to skip past
+#  the expansion animation that occurs at start_time every hour. can probably do this by adding or subtracting
+#  1/expansion somewhere
+
+#interface for the instance uniforms is 
+#INSTANCE_CUSTOM (COLOR).x = type selection, .y = start time
+
+func add_selection_marker(unit_id : int, unit_position : Vector3) -> void:
```
Hop311, line 42: This function doesn't check if there's an existing instance with ID `unit_id`, which could lead to issues when trying to update selections later. 
Also we'll need a way of removing specific selection markers

Hop311 (dismissed):

game/src/Game/GameSession/SelectionMarkers.gd
Nemrav: You can have either colour, or custom or both. We are only using custom, so comes to 16 in total.

game/src/Game/GameSession/SelectionMarkers.gd
Nemrav: Technically what I've implemented is a soft-cap, but I suppose we could also check that.

game/src/Game/GameSession/SelectionMarkers.gd
Nemrav: This does pack the data more efficiently, but I think spending time zipping the ids and positions together might waste any gains you'd get?

game/src/Game/GameSession/MarkerManager.gd
```diff
@@ -0,0 +1,122 @@
...
+		var duration : float = projection[duration_key]
+		var additative : bool = projection[additative_key]
+
+		#fix the alpha edges of the projection textures
+		var texture : ImageTexture = AssetManager.get_texture(texture_name)
+		if texture == null:
+			push_error("Texture for projection \"", projection_name, "\" was null!")
+			continue
+		var image : Image = texture.get_image()
+		image.fix_alpha_edges()
```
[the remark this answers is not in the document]
Nemrav: All I know about it are what's said in the docs:
- https://docs.godotengine.org/en/stable/tutorials/assets_pipeline/importing_images.html#process-fix-alpha-border
- https://docs.godotengine.org/en/stable/classes/class_image.html#class-image-method-fix-alpha-edges

From the description, I think we want to do it as early as possible for all images with an alpha channel.
I'll experiment with adding
```
	if(image->detect_alpha() != Image::ALPHA_NONE){
		image->fix_alpha_edges();
	}
```
To AssetManager's _load_image function.

game/src/Game/GameSession/SelectionMarkers.gd
Nemrav: An unnecessary left-over, I am now using .z as a transparency override for selection marker deletion.

game/src/Game/GameSession/MarkerManager.gd
Nemrav: It isn't necessary, removed.

extension/src/openvic-extension/singletons/MapItemSingleton.hpp
```diff
@@ -25,14 +24,23 @@ namespace OpenVic {
 
 	private:
 		GFX::Billboard const* get_billboard(std::string_view name, bool error_on_fail = true) const;
-		bool add_billboard_dict(std::string_view name, godot::TypedArray<godot::Dictionary>& billboard_dict_array) const;
+		void add_billboard_dict(GFX::Billboard const& billboard, godot::TypedArray<godot::Dictionary>& billboard_dict_array) const;
 		godot::TypedArray<godot::Dictionary> get_billboards() const;
+
+		GFX::Projection const* get_projection(std::string_view name, bool error_on_fail = true) const;
```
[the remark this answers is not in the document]
Nemrav: Removed those functions.

extension/src/openvic-extension/singletons/MapItemSingleton.cpp
```diff
@@ -94,8 +94,64 @@ TypedArray<Dictionary> MapItemSingleton::get_billboards() const {
...
+		}
+	}
+
+	return ret;
+}
+
+
+GFX::Projection const* MapItemSingleton::get_projection(std::string_view name, bool error_on_fail) const {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton();
+	ERR_FAIL_NULL_V(game_singleton, nullptr);
```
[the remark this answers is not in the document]
Nemrav: Removed the game_singleton checks from the file. I've left in the definition manager checks though.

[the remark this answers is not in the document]
Hop311: I think you mean `InstanceManager`, and yes those should stay as it could be null if someone calls these functions while there is no active game session (e.g. calling them from GDScript while on the main main)

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,144 @@
...
+	)
+	return true
+
+#TODO: Perhaps there's a more efficient method...
+func add_selection_markers(unit_ids : PackedInt32Array, unit_positions : PackedVector3Array) -> bool:
+	assert(unit_ids.size() == unit_positions.size())
+	# don't increase visible count all at once here, in-case an id is invalid
+	var ret : bool = true
+	for index : int in unit_ids.size():
+		ret = ret and add_selection_marker(unit_ids[index],unit_positions[index])
```
Hop311, line 93:
```suggestion
		if not add_selection_marker(unit_ids[index],unit_positions[index]):
			ret = false
```
In C++ code we use the bitwise `&` operator as it ensures the second expression is executed, rather than the logical `&&` operator which won't execute the second expression if the first evaluates to false.
GDScript's `and` has the same issue, so we also can't use it here. Might as well use an `if` statement here, the performance gain of using `&` would be negligible for an interpreted language like GDScript

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,144 @@
...
+	return ret
+
+func update_selection_markers(unit_ids : PackedInt32Array, unit_positions : PackedVector3Array) -> bool:
+	assert(unit_ids.size() == unit_positions.size())
+	var ret : bool = true
+	for index : int in unit_ids.size():
+		ret = ret and update_selection_marker(unit_ids[index],unit_positions[index])
```
Hop311, line 100:
```suggestion
		if not update_selection_marker(unit_ids[index],unit_positions[index]):
			ret = false
```

game/src/Game/GameSession/SelectionMarkers.gd
```diff
@@ -0,0 +1,144 @@
...
+	for i : int in unused_indices.size():
+		unused_indices[i] = i
+	id_to_index.clear()
+	multimesh.visible_instance_count = 0
+
+func remove_selection_marker(unit_id : int) -> bool:
+	var index : int = id_to_index.get(unit_id,-1)
+	if index == -1:
+		return false # doesn't have that id!
+	unused_indices[unused_indices.find(-1)] = index
```
Hop311, line 119: Are we always sure there'll be an entry with value -1 available to find here? I guess so if there are always at least as many entries in `unused_indices` as in `id_to_index`

Hop311 (commented):
Not essential, but it would be nice if shift clicking a province or port that already has a selection marker removes it as an example of using the `remove_selection_marker` function.

Hop311 (approved):
Amazing work, thanks Nemrav 😄

PR #336, opened 2024-12-23
Add pre-commit configuration file

Spartan322:
Add typos checks and corrections
Add static check github action
Add clang-tidy corrections
Use optional initialization for PROPERTY macros

Implements everything but clang-format from #305

When this is merged, the `victoria 2 discrepency` needs to immediately be renamed to `victoria 2 discrepancy`


docs/simulation/calculations.yaml
```diff
@@ -7,7 +7,7 @@ Diplomacy:
       else: Total base influence gain / number of countries (excluding banned)
     Country influence gain modifier:
       note: 'our' and 'we' refers to the influencer. 'target' is the country being influenced.
-      formula: 1 + score mod + relations mod + population mod + investment mod + puppet mod + neighbour mod + continent mod + discredit mod
+      formula: 1 + score mod + relations mod + population mod + investment mod + puppet mod + neighbor mod + continent mod + discredit mod
```
wvpm, line 10: Aren't we using British English?

wvpm (approved):

wvpm (dismissed):

wvpm (dismissed):

extension/src/openvic-extension/classes/GUIListBox.cpp
```diff
@@ -128,9 +128,7 @@ void GUIListBox::_notification(int what) {
 	}
 }
 
-GUIListBox::GUIListBox()
-  : gui_listbox { nullptr }, scrollbar { nullptr }, scroll_index { 0 }, max_scroll_index { 0 },
-	fixed { false }, fixed_item_count { 0 }, fixed_visible_items { 0 }, fixed_item_height { 0.0_real } {}
```
Hop311, line 133: `scroll_index`, `max_scroll_index`, `fixed`, `fixed_item_count` and `fixed_visible_items` all need to be initialised

Hop311 (commented):
A few primitives in `GUIListBox` have had their constructor initialisation removed with no replacement in the header.
Also if we need to move initialisation into the header/declaration, it would be better to do so completely rather than leaving a mix between the two which makes it difficult to tell what is and isn't initialised.

Hop311 (dismissed):

PR #348, opened 2025-01-19
Add military menu control script

Hop311:


extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -110,58 +111,126 @@ String MenuSingleton::get_country_adjective(CountryInstance const& country) cons
...
+) {
+	String result;
+
+	if (plus_for_non_negative && value >= 0) {
+		result = "+";
+	}
+
+	using enum ModifierEffect::format_t;
+
+	switch (format_effect.get_format()) {
```
wvpm, line 127: Why not move this switch to the modifier effect?
The menu singleton already has enough things to do.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -872,6 +1006,803 @@ String MenuSingleton::get_longform_date() const {
...
+		leader_dict[military_info_leader_name_key] = std::move(leader_name);
+	}
+
+	{
+		// Prestige
+		const fixed_point_t prestige = leader.get_prestige();
+		fixed_point_t morale_bonus, organisation_bonus;
+
+		GameSingleton const* game_singleton = GameSingleton::get_singleton();
+		if (game_singleton != nullptr) {
```
wvpm, line 1114: Game singleton is never null, right?

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -872,6 +1006,803 @@ String MenuSingleton::get_longform_date() const {
...
+		String prestige_tooltip = tr(prestige_localisation_key).replace(
+			value_replace_key, Utilities::float_to_string_dp(prestige * 100, 2) + "%"
+		);
+
+		tooltip += "\n" + prestige_tooltip;
+
+		static const StringName morale_localisation_key = "PRESTIGE_MORALE_BONUS";
+		static const StringName organisation_localisation_key = "PRESTIGE_MAX_ORG_BONUS";
+
+		// Morale and organisation bonuses are always green with a + sign, matching the base game's behaviour
```
wvpm, line 1138: When negative they shouldn't be green.
Also can't you reuse the code to format modifier effects here?

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -872,6 +1006,803 @@ String MenuSingleton::get_longform_date() const {
...
+	unit_group_dict[military_info_unit_group_name_key] = Utilities::std_to_godot_string(unit_group.get_name());
+	if (unit_group.get_position() != nullptr) {
+		unit_group_dict[military_info_unit_group_location_key] =
+			Utilities::std_to_godot_string(unit_group.get_position()->get_identifier());
+	}
+	unit_group_dict[military_info_unit_group_unit_count_key] = static_cast<uint64_t>(unit_group.get_unit_count());
+
+	if constexpr (Branch == LAND) {
+		// TODO - calculate (max) men and morale properly in ArmyInstance and RegimentInstances and set here
+		unit_group_dict[military_info_unit_group_men_count_key] = static_cast<uint64_t>(unit_group.get_unit_count() * 3000);
```
wvpm, line 1244: What are the 3000's doing here?
This should not be hardcoded.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -872,6 +1006,803 @@ String MenuSingleton::get_longform_date() const {
...
+		).replace(moving_date_replace_key, Utilities::date_to_string(arrival_date));
+	}
+
+	if constexpr (Branch == LAND) {
+		// TODO - check if digging in, and if true get days spent digging in
+		if (true /* unit_group.is_digging_in() */) {
+			static const StringName digin_localisation_key = "MILITARY_DIGIN_TOOLTIP";
+			static const String moving_days_replace_key = "$DAYS$";
+
+			int64_t days_spent_digging_in = 4;
```
wvpm, line 1274: Again, this should not be hardcoded.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: We can't move it to the `ModifierEffect` as it uses Godot `String`s which aren't accessible in the SIM submodule. It could indeed be worth moving the modifier tooltip functions out of `MenuSingleton`, but we couldn't just make a helper namespace full of static functions like `Utilities`, we need a non-static Godot `Object` instance to provide access to the translation `tr` function, so basically another singleton of some kind.
For now I think it's simplest to keep it here, I have 2 other PRs building on top of this and Brick has one too, we can rearrange stuff in a later cleanup PR (tbh I might keep the modifier stuff here and move other stuff from this file, we already have a separate source file for `MenuSingleton`'s population menu functions, and I've been considering reworking our menu system to replace all these data packaging functions with custom C++ menu nodes directly updating their UI child nodes with data read from the SIM)

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: Yes, I recently realised that it can't be null unless the whole program has failed catastrophically. I'm going to remove all these singleton null checks in an upcoming PR, there are loads across many files and it'll be simpler to change them in one go rather than changing these ones now and having to rebase all the other PRs dependent on this PR.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: These are examples so reviewers can see that the UI side works, we don't yet have the SIM side data but I thought better to have something rather than leaving it completely unimplemented.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: Yes, I recently realised that it can't be null unless the whole program has failed catastrophically. I'll remove the ones currently added by this PR, as for the pre-existing ones I'll remove them all in an upcoming cleanup PR.

wvpm (approved):

wvpm (approved):

PR #349, opened 2025-01-20
Starting bookmark position, search panel enter, cleanup

Hop311:
Also fixes some model attachment errors and brings in SIM changes that fix duplicate core errors

### DIFF
--- a/extension/doc_classes/GameSingleton.xml
+++ b/extension/doc_classes/GameSingleton.xml
@@ -3,188 +3,247 @@
 	<brief_description>
 	</brief_description>
 	<description>
+		This is the GameSingleton
 	</description>
 	<tutorials>
 	</tutorials>
 	<methods>
+		<method name="get_bookmark_info" qualifiers="const">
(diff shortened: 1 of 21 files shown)


extension/doc_classes/GameSingleton.xml
```diff
@@ -7,6 +7,16 @@
 	<tutorials>
 	</tutorials>
 	<methods>
+		<method name="get_bookmark_info" qualifiers="const">
+			<return type="Dictionary[]" />
+			<description>
+			</description>
```
Nemrav, line 13: Could you add descriptions to these new functions for documentation. Something short like "Returns a dictionary with bookmark info, key `x` gives a String ..., key `y` gives an int ..."

Nemrav (approved):
This is all really good, brings us closer to having a fully recreated lobby. Just add the documentation and change the camera to zoom to the capital instead of the bookmark location and I think its good to go.

Nemrav (approved):

PR #355, opened 2025-01-26
Add/improve RGO production and size tooltips

Hop311:


extension/src/openvic-extension/classes/GUINode.hpp
```diff
@@ -88,11 +88,14 @@ namespace OpenVic {
...
 		static godot::String int_to_string_suffixed(int64_t val);
+		static godot::String int_to_string_commas(int64_t val);
 		static godot::String float_to_string_suffixed(float val);
 		static godot::String float_to_string_dp(float val, int32_t decimal_places);
 		// 3dp if abs(val) < 2 else 2dp if abs(val) < 10 else 1dp
 		static godot::String float_to_string_dp_dynamic(float val);
-		static godot::String format_province_name(godot::String const& province_identifier);
+		// The "ignore_empty" argument refers to what this function produces when given an empty string - if the argument
+		// is false then empty inputs are replaced with "NO PROVINCE", otherwise they return the empty string unchanged.
+		static godot::String format_province_name(godot::String const& province_identifier, bool ignore_empty = false);
```
wvpm, line 98: What are the usecases for true and false here?
It seems to me false is only useful for development to spot the "NO PROVINCE" texts.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -187,6 +254,57 @@ String MenuSingleton::make_rules_tooltip(RuleSet const& rules) const {
 	return result;
 }
 
+String MenuSingleton::_make_mobilisation_impact_tooltip() const {
```
wvpm, line 257: Shouldn't this be in the military menu PR?

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -459,68 +598,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
-	ret[province_info_rgo_total_employees_key] = rgo.get_total_employees_count_cache();
-	const pop_size_t max_employee_count = rgo.get_max_employee_count_cache();
-	if (max_employee_count == 0) {
-		ret[province_info_rgo_employment_percentage_key] = 100.0f;
-	} else {
-		ret[province_info_rgo_employment_percentage_key] =
-			(rgo.get_total_employees_count_cache() * fixed_point_t::_100() / max_employee_count).to_float_rounded();
-	}
 
 	if (rgo.is_valid()) {
```
wvpm, line 635: Invert this so `!rgo.is_valid()` is handled first.
The `rgo.is_valid()` case is so long you forget you're in an if statement reading it.
Probably a good idea to move the rgo code to a new method.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -459,68 +598,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+		fixed_point_t output_from_workers = fixed_point_t::_1(), throughput_from_workers = fixed_point_t::_0(),
+			output_multiplier = fixed_point_t::_1(), throughput_multiplier = fixed_point_t::_1();
+		String size_string, output_string, throughput_string;
+
+		static const String employee_effect_template_string = "\n  -%s: " + GUILabel::get_colour_marker() + "Y%s" +
+			GUILabel::get_colour_marker() + "!: %s";
+
+		using enum Job::effect_t;
+
+		std::optional<Job> const& owner = production_type.get_owner();
```
wvpm, line 664: `owner_job`?
To prevent confusion with province owner.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -459,68 +598,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+					employees_of_type
+				);
+			}
+
+			for (Job const& job : production_type.get_jobs()) {
+				if (job.get_pop_type() != &pop_type) {
+					continue;
+				}
+
+				// TODO - exclude pop types with no members present in state/province at all??? e.g. slaves
```
wvpm, line 727: I don't understand this reasoning.
You're looping through the `employee_count_per_type_cache`.
This only contains pops of a type that are in the state and have a job in the production type.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -459,68 +598,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+				);
+				output_modifiers += _make_modifier_effect_contributions_tooltip(
+					*province, *modifier_effect_cache.get_farm_rgo_output_local(), &output_from_tech, &output_multiplier,
+					modifier_effect_contributions_prefix
+				);
+
+				if (terrain_type != nullptr) {
+					size_from_terrain = terrain_type->get_effect(*modifier_effect_cache.get_farm_rgo_size_local());
+				}
+				size_from_province = province->get_modifier_effect_value(*modifier_effect_cache.get_farm_rgo_size_local());
```
wvpm, line 805: Does this not include the terrain effect?

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -459,68 +598,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+			ModifierEffectCache::good_effects_t const& good_effects =
+				modifier_effect_cache.get_good_effects()[production_type.get_output_good()];
+
+			output_from_tech += province->get_modifier_effect_value(*good_effects.get_rgo_goods_output());
+			throughput_from_tech += province->get_modifier_effect_value(*good_effects.get_rgo_goods_throughput());
+
+			fixed_point_t size_from_tech;
+
+			size_from_tech = province->get_modifier_effect_value(*good_effects.get_rgo_size());
+			if (production_type.get_is_farm_for_tech()) {
```
wvpm, line 856: I expected this above the non-tech variant as you start off with `output_from_tech` & `throughput_from_tech`.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
wvpm, line 1: I miss the code adding tooltips for each event modifier, issue & NF impact on RGOs.
They are listed individually instead of grouped as tech.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: I had forgotten to update this branch after modifying the military-menu branch, so all the changes from that old military-menu commit were listed in the changes tab for this PR. I've since updated this branch and fixed the issue

extension/src/openvic-extension/classes/GUINode.hpp
Hop311: In some places an empty province identifier is an error, e.g. the title of the province overview panel, and so we want to be made aware that something has gone wrong there.
In other places (specifically the location label on leader entries in the military menu) empty identifiers are allowed, so we let them through without appending `"PROV"` and without replacing them with `"NO PROVINCE"`.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: This is an `IndexedMap` so it contains an entry for every pop type (we may want to change to an `ordered_map<PopType const*, pop_size_t>` instead as we are wasting a lot of space on entries that are always 0).
I was unsure how to handle pop types with jobs but without any pops present in the state, as owners are listed in the tooltip even if there are none of them and their contribution is 0%. I decided to follow owners' behaviour and always list all pop types with jobs, but then I realised this was wrong as it resulted in `Slaves` being listed in RGO production tooltips which don't have it in the base game. Upon further testing, I realised that only owners are always shown, while regular employees are only shown if any of them are actually employed at that RGO.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: It does include the terrain effect, the tooltip lists the terrain contribution as well as the total province contribution including terrain.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: Those are generated by calls to `_make_modifier_effect_contributions_tooltip`, a helper function which creates tooltip lines for every source contributing to a resultant effect, with tech contributions optionally skipped if a fixed point pointer is provided for them to be added to instead.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: `output_from_tech` & `throughput_from_tech` are declared at the top because they're used as arguments in the `_make_modifier_effect_contributions_tooltip` at the top of this block, which adds tech contributions to various modfiier effects to those variables. The section you commented on goes here because I grouped everything dependent on `get_is_farm_for_tech()` in one place so we don't need multiple identical if statements, and these if statements go after the `get_is_farm_for_non_tech()` if statements beacuse the non-tech contributions appear first in the tooltip

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: There is no `!rgo.is_valid()` case, we don't want to print errors every time a province with no RGO is selected (currently that includes water provinces which of course won't be selectable eventually, but even in the base game you can have provinces with no RGO if you don't provide a `trade_goods = ...` line in their province history file)

wvpm: In that case, it's best to move this to a separate method for readability.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -603,68 +605,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+			GUILabel::get_colour_marker() + "!: %s";
+
+		using enum Job::effect_t;
+
+		std::optional<Job> const& owner_job = production_type.get_owner();
+		if (owner_job.has_value()) {
+			PopType const& owner_pop_type = *owner_job->get_pop_type();
+			State const* state = province->get_state();
+
+			if (state != nullptr) {
```
wvpm, line 676: Handled error cases first.
I'll try to explain reading experience:
```
if (condition) { //reader understands condition and enters the if
long code block //reader is focused on all the stuff happening in here
} else { //reader is somewhat surprised we're still in the if and may need to look up condition again
error //oh nothing special
}
```

vs 
```
if (condition) {
error //ah it's just an error check
} else { //forget the condition, we're past it
long block of code //reader can focus on long block of code
}
```

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -603,68 +605,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+
 		String amount_of_employees_by_pop_type;
+
 		for (auto const& [pop_type, employees_of_type] : rgo.get_employee_count_per_type_cache()) {
-			if (employees_of_type > 0) {
-				amount_of_employees_by_pop_type +=
-					"  -" + GUILabel::get_colour_marker() + "Y" +
-					tr(Utilities::std_to_godot_string(pop_type.get_identifier())) + GUILabel::get_colour_marker() + "!:" +
-					String::num_int64(employees_of_type) + "\n";
+			if (employees_of_type <= 0) {
```
wvpm, line 719: Nice early exit.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -603,68 +605,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+				output_modifiers += _make_modifier_effect_contributions_tooltip(
+					*province, *modifier_effect_cache.get_farm_rgo_output_local(), &output_from_tech, &output_multiplier,
+					modifier_effect_contributions_prefix
+				);
+
+				if (terrain_type != nullptr) {
+					size_from_terrain = terrain_type->get_effect(*modifier_effect_cache.get_farm_rgo_size_local());
+				}
+				size_from_province = province->get_modifier_effect_value(*modifier_effect_cache.get_farm_rgo_size_local());
+			}
```
wvpm, line 813: Unrelated if statements should have a new line in between to make clear they are unrelated.
Otherwise they look like an else if when glancing through the code.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -603,68 +605,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+
+			size_from_tech = province->get_modifier_effect_value(*good_effects.get_rgo_size());
+			if (production_type.get_is_farm_for_tech()) {
+				const fixed_point_t value =
+					province->get_modifier_effect_value(*modifier_effect_cache.get_farm_rgo_throughput_and_output());
+				output_from_tech += value;
+				throughput_from_tech += value;
+
+				size_from_tech += province->get_modifier_effect_value(*modifier_effect_cache.get_farm_rgo_size_global());
+			}
```
wvpm, line 870: Again unrelated ifs without room in between

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -603,68 +605,379 @@ Dictionary MenuSingleton::get_province_info_from_index(int32_t index) const {
...
+				static const StringName rgo_output_tech_localisation_key = "RGO_OUTPUT_TECH";
+
+				output_string += vformat(
+					tech_modifier_template_string,
+					tr(rgo_output_tech_localisation_key),
+					_make_modifier_effect_value_coloured(
+						*modifier_effect_cache.get_rgo_output(), output_from_tech, true
+					)
+				);
+			}
```
wvpm, line 902: ditto

wvpm (approved):

PR #360, opened 2025-01-28
Add research and leadership topbar tooltips and start research example

Hop311:


extension/src/openvic-extension/singletons/GameSingleton.cpp
```diff
@@ -174,9 +174,22 @@ Error GameSingleton::setup_game(int32_t bookmark_index) {
 	ret &= menu_singleton->_population_menu_update_provinces() == OK;
 
 	// TODO - replace with actual starting country
-	set_viewed_country(instance_manager->get_country_instance_manager().get_country_instance_by_identifier("ENG"));
+	CountryInstance* starting_country =
```
wvpm, line 177: Since the code will crash if it is null, just use a reference.

extension/src/openvic-extension/singletons/MenuSingleton.cpp
```diff
@@ -925,6 +1053,77 @@ Dictionary MenuSingleton::get_topbar_info() const {
...
+			leadership_tooltip += tr(pop_type_leadership_localisation_key).replace(
+				pop_type_replace_key, tr(Utilities::std_to_godot_string(pop_type->get_identifier()))
+			).replace(
+				value_replace_key, Utilities::fixed_point_to_string_dp(leadership_points, 2)
+			).replace(
+				fraction_replace_key, Utilities::fixed_point_to_string_dp(
+					100 * country->get_pop_type_proportion(*pop_type) / country->get_total_population(), 2
+				)
+			).replace(
+				optimal_replace_key, Utilities::fixed_point_to_string_dp(100 * pop_type->get_research_leadership_optimum(), 2)
```
wvpm, line 1085: As percentages tend to be displayed with 2 decimals, this could be a utility method.
`Utilities::fixed_point_to_string_dp(100 * fractional_value, 2)`

wvpm (approved):

extension/src/openvic-extension/singletons/GameSingleton.cpp
Hop311: It won't crash, it will print some errors and then start the game with no viewed country selected, same as when you right click on a province with no owner

extension/src/openvic-extension/singletons/MenuSingleton.cpp
Hop311: Cool idea, might add it in a future PR

PR #361, opened 2025-01-28
Add support for inline flags in GUILabel

Hop311:


extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -628,6 +645,78 @@ void GUILabel::separate_lines(
...
+		}
+	} else {
+		country_definition =
+			game_singleton->get_definition_manager().get_country_definition_manager().get_country_definition_by_identifier(
+				Utilities::godot_to_std_string(identifier)
+			);
+		ERR_FAIL_NULL_V(country_definition, {});
+	}
+
+	// TODO - if country is null, fallback to REB/country index 0 and empty flag type
```
wvpm, line 701: Rebel flag is not used for provinces without owner. Rebel flag is pretty rare tbh.

extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -628,6 +645,78 @@ void GUILabel::separate_lines(
...
+		start_pos = marker_pos + get_currency_marker().length();
+	}
+
+	if (start_pos < string.length()) {
+		separate_flag_segments(string.substr(start_pos), colour, line);
+	}
+}
+
+GUILabel::flag_segment_t GUILabel::make_flag_segment(String const& identifier) {
+	GameSingleton const* game_singleton = GameSingleton::get_singleton(); // Should never be null
```
wvpm, line 671: Use references instead of pointers if we are certain it has a value.

extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -150,30 +161,34 @@ void GUILabel::_notification(int what) {
...
-					if (currency_texture.is_valid()) {
-						currency_texture->draw(
-							ci, position - Vector2 {
-								1.0_real, static_cast<real_t>(currency_texture->get_height()) * 0.75_real
-							}
-						);
-						position.x += currency_texture->get_width();
-					}
-				} else {
+				if (string_segment_t const* string_segment = std::get_if<string_segment_t>(&segment)) {
```
wvpm, line 167: Instead of making if else structure here, could the code be moved to segment and use a shared interface?
`segment.draw()` something like that.

wvpm (dismissed):

extension/src/openvic-extension/classes/GUILabel.cpp
Hop311: Country being null here has nothing to do with optional country variables like province owner not being set, it's about the result of a `get_country_*_by_identifier` lookup failing, i.e. no country existing with the given tag. The base game handles these cases by falling back to the rebel flag, you can test this by adding `@ABC` or `@123` to a localisation string and looking at the result in-game. I believe it specifically falls back to the first flag in the `flagfiles.tga` spritesheet (located in `gfx/flags` under the same directory as `save games`), you can test this by changing which country comes first in `common/countries.txt` and deleting the flag spritesheet then starting the game and looking at the localisation of `@123` once again.
I'll recreate this behaviour by falling back to the country index `0` and empty flag type when a country cannot be found for the given identifier, this should always return the first defined tag's flag (unless there are no countries at all in which case I'll just return `nullptr`).

extension/src/openvic-extension/classes/GUILabel.cpp
Hop311: `segment_t` is a variant, not a custom class we've made. The best we could do is to separate this out into another function, still as a `if`/`else if`/`else` tower, but given how integrated the code inside the `for` loop is with the rest of this function's code, we'd need a mess of arguments for all the local variables which would obfuscate their source and how they evolve over time.
The `position` variable in particular has very important and specific logic necessary to draw all the different lines and kinds of segments in the correct places relative to each other, and having all the code involved in that logic close together is very useful when modifying it.

wvpm (approved):

PR #363, opened 2025-01-29
Add GUILineChart with examples in budget and trade charts

Hop311:


extension/src/openvic-extension/classes/GUILineChart.cpp
```diff
@@ -0,0 +1,239 @@
...
+	);
+
+	return set_gfx_line_chart(new_gfx_line_chart);
+}
+
+String GUILineChart::get_gfx_line_chart_name() const {
+	return gfx_line_chart != nullptr ? Utilities::std_to_godot_string(gfx_line_chart->get_name()) : String {};
+}
+
+Error GUILineChart::set_gradient_line(PackedFloat32Array const& line_values, float value_range) {
```
wvpm, line 88: Why would you provide value_range if the method calculates it for you?
Providing it does not skip the calculation part.

wvpm (approved):

extension/src/openvic-extension/classes/GUILineChart.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+	Ref<Gradient> gradient;
+	gradient.instantiate();
+	ERR_FAIL_NULL_V(gradient, FAILED);
+
+	// Remove hardcoded (1, WHITE) point
+	gradient->remove_point(1);
+
+	// Calculate the height of the graph in value-space
+	for (int64_t index = 0; index < line_values.size(); ++index) {
```
wvpm, line 112: I still don't understand why this loop is here if `min_value_range` is provided.
Surely the caller has a similar loop to figure out what to pass to this method.

extension/src/openvic-extension/classes/GUILineChart.cpp
Hop311: It lets you specify a minimum/starting value, so that the top and bottom of the graph represent at least +/- value_range. I've renamed it to `min_value_range` in the recent force push.

extension/src/openvic-extension/classes/GUILineChart.cpp
Hop311: `min_value_range` is a minimum, so that the top/bottom values represent at least +/- min_value_range. We still do the loop in case we need to go even further out, as we don't expect `min_value_range` to be set based on the caller calculating it from the values but rather using a constant depending on which menu the graph is in. For example, in the trade menu the graphs range is at least `base price +/- £1`, it can be larger but not smaller, and that would be specified with a `min_value_range` of `1`.

wvpm: Can you change it to padding then?

wvpm (approved):

PR #367, opened 2025-02-02
Add trade menu control script

Hop311:


extension/src/openvic-extension/classes/GUILineChart.cpp
```diff
@@ -117,6 +120,9 @@ Error GUILineChart::set_gradient_line(PackedFloat32Array const& line_values, flo
 		}
 	}
 
+	min_value = central_value - min_value_range;
```
wvpm, line 123: Why is this executed before updating min_value_range?

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+	Dictionary ret;
+
+	for (GoodCategory const& good_category : good_definition_manager.get_good_categories()) {
+		TypedArray<Dictionary> array;
+
+		for (GoodDefinition const* good_definition : good_category.get_good_definitions()) {
+			GoodInstance const& good_instance = good_instance_manager.get_good_instance_from_definition(*good_definition);
+
+			if (good_instance.get_is_available() && good_definition->get_is_tradeable()) {
```
wvpm, line 38: Move to method as you're already 2 for loops deep.
```
for (GoodCategory const& good_category : good_definition_manager.get_good_categories()) {
for (GoodDefinition const* good_definition : good_category.get_good_definitions()) {
do(good_category, good_definition)
}
}
```

Reads way better.

wvpm: The experience reading
https://i.etsystatic.com/35826945/r/il/cacfd7/5178567711/il_fullxfull.5178567711_i90p.jpg

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -0,0 +1,312 @@
...
+						val_replace_key, Utilities::fixed_point_to_string_dp(demand, 3)
+					) + "\n" + tr(actual_bought_localisation_key).replace(
+						val_replace_key, Utilities::fixed_point_to_string_dp(good_instance.get_quantity_traded_yesterday(), 3)
+					);
+				}
+
+				if (country != nullptr) {
+					CountryInstance::good_data_t const& good_data = country->get_goods_data()[good_instance];
+
+					int32_t trade_settings = 0;
```
wvpm, line 70: I don't understand this part of the code at all.
Why multiply and binary or?
What are the desired outcomes?

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -0,0 +1,312 @@
...
+			ret[trade_detail_price_history_key] = std::move(price_history);
+		} else {
+			UtilityFunctions::push_error(
+				"Failed to resize price history array to the correct size (",
+				static_cast<int64_t>(good_price_history.size()), ")"
+			);
+		}
+	}
+
+	if (country != nullptr ) {
```
wvpm, line 153: How can county even be null inside the trade menu?
Try to reduce nesting when possible.
```
if (OV_unlikely(country == nullptr)) {
//error
return ret;
}

//rest of code

File: extension/src/openvic-extension/singletons/TradeMenu.cpp [hunk 5]
```diff
@@ -0,0 +1,312 @@
...
+	for (GoodInstance const& good : good_instance_manager.get_good_instances()) {
+		static const StringName top_producers_localisation_key = "TRADE_TOP_PRODUCERS";
+
+		String tooltip = tr(Utilities::std_to_godot_string(good.get_identifier())) + get_tooltip_separator() +
+			tr(top_producers_localisation_key);
+
+		for (size_t index = 0; index < 5; ++index) {
+			CountryInstance const* country = country_instance_manager.get_country_instance_by_index(index + 1);
+
+			if (country != nullptr) {
```
wvpm, line 206:
```
if (country == nullptr) {
continue;
}
```

Try to reduce nesting. For the love of readability.

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -0,0 +1,312 @@
...
+		PackedVector2Array government_needs;
+		PackedVector2Array factory_needs;
+		PackedVector2Array pop_needs;
+		PackedVector3Array market_activity;
+		PackedVector3Array stockpile;
+		PackedVector4Array common_market;
+
+		for (auto const& [good, good_data] : country->get_goods_data()) {
+			if (!good.get_is_available() || !good.get_good_definition().get_is_tradeable()) {
+				continue;
```
wvpm, line 233: Hurray!

game/src/Game/GameSession/NationManagementScreen/TradeMenu.gd
```diff
@@ -46,7 +226,374 @@ func _on_update_active_nation_management_screen(active_screen : NationManagement
...
+		(func(a : Node, b : Node) -> bool: return a.get_meta(sort_key) > b.get_meta(sort_key))
+		if descending else
+		(func(a : Node, b : Node) -> bool: return a.get_meta(sort_key) < b.get_meta(sort_key))
+	)
+
+	for child : Node in items:
+		listbox.add_child(child)
+
+func _float_to_string_suffixed_dp(value : float, decimals : int) -> String:
+	if value < 1000:
```
wvpm, line 416: Shouldn't this be < 10000?
Otherwise you don't get any benefit.
"1.234K".length > "1234".length

Hop311: This is matching the base game's behaviour where, in certain trade menu table columns, values >= 1000 get divided by 1000 with the "k" suffix added.
Also it wouldn't be "1.234k" vs "1234", it would be "1.234k" vs "1234.000" as it uses a fixed number of decimal places, not significant figures.

extension/src/openvic-extension/singletons/TradeMenu.cpp
Hop311: `country` here refers to the currently selected player/viewed country, which can be null if you right click on a province with no owner. This will eventually become some kind of "observer mode", but for now we just skip parts which require a specific country

extension/src/openvic-extension/singletons/TradeMenu.cpp
Hop311: I'll change the availability check to continue when not available, but create a separate function for the inner section would if anything reduce readability IMO with the need to pass the variables uses across multiple iterations as reference arguments.

extension/src/openvic-extension/singletons/TradeMenu.cpp
Hop311: I'll make it use bitflag enums and add a comment explaining what they represent

extension/doc_classes/MenuSingleton.xml
```diff
@@ -337,6 +353,14 @@
 		</constant>
 		<constant name="SORT_LITERACY" value="17" enum="PopSortKey">
 		</constant>
+		<constant name="TRADE_SETTING_NONE" value="0" enum="TradeSettingBit">
```
wvpm, line 356: So the values used would be:
2, manual buying
1+2=3, automated buying
4, manual selling

Feel weird both 0 and 1 (and > 4) are invalid.

Hop311: You can have 1, so automated but with no buying or selling icon (this exclusively refers to the icons on the goods in the big grid)

wvpm (approved):

PR #368, opened 2025-02-02
Fix nullptr crash in `GUILabel::generate_lines_and_segments`

Spartan322:


extension/src/openvic-extension/classes/GUILabel.cpp
```diff
@@ -584,16 +585,19 @@ std::vector<GUILabel::line_t> GUILabel::generate_lines_and_segments(
 
 	for (int64_t idx = 0; idx < display_text.length(); ++idx) {
 		Color new_colour = current_colour;
-		while (colour_it != colour_instructions.end() && idx == colour_it->first) {
+		for (; colour_it != colour_instructions.end(); colour_it++) {
```
Hop311, line 588:
```suggestion
		for (; colour_it != colour_instructions.end() && idx == colour_it->first; colour_it++) {
```
The index check needs to go here rather than at the bottom, otherwise colours will be applied early rather than waiting for their intended position. If you run the game with this PR you'll see labels like the topbar's budget and population text aren't being formatted correctly.

Hop311 (approved):

PR #376, opened 2025-02-05
Make PlayerSingleton manage game actions via new SIM action queue system

Hop311:
Will come out of draft once [OpenVic-Simulation#296](https://github.com/OpenVicProject/OpenVic-Simulation/pull/296) is merged

### DIFF
--- a/extension/src/openvic-extension/singletons/PlayerSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/PlayerSingleton.cpp
@@ -3,7 +3,9 @@
 #include <openvic-simulation/country/CountryInstance.hpp>
 #include <openvic-simulation/map/ProvinceInstance.hpp>
 
+#include "openvic-extension/classes/GUIScrollbar.hpp"
 #include "openvic-extension/singletons/GameSingleton.hpp"
+#include "openvic-extension/singletons/MenuSingleton.hpp"
 #include "openvic-extension/utility/ClassBindings.hpp"
 
 using namespace OpenVic;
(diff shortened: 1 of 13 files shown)


extension/src/openvic-extension/singletons/PlayerSingleton.cpp
```diff
@@ -44,14 +75,28 @@ PlayerSingleton::~PlayerSingleton() {
 }
 
 // Player country
-void PlayerSingleton::set_player_country(CountryInstance* new_player_country) {
+void PlayerSingleton::set_player_country(CountryInstance const* new_player_country) {
```
wvpm, line 78: Nice to see the instances being const now.

extension/src/openvic-extension/singletons/PlayerSingleton.cpp
```diff
@@ -44,14 +75,28 @@ PlayerSingleton::~PlayerSingleton() {
...
 	if (player_country != new_player_country) {
```
wvpm, line 79: Invert if to return immediately.

extension/src/openvic-extension/singletons/PlayerSingleton.cpp
```diff
@@ -118,3 +163,144 @@ void PlayerSingleton::unset_selected_province() {
...
+			player_country->get_index(), good_index, is_selling,
+			MenuSingleton::calculate_trade_menu_stockpile_cutoff_amount_fp(amount_slider->get_value_scaled_fp())
+		}
+	);
+}
+
+// Diplomacy
+
+// Military
+void PlayerSingleton::create_leader(bool is_general) const {
```
wvpm, line 250: I know from just having reviewed the sim PR that false means it's navy but this feels confusing if you haven't.

extension/src/openvic-extension/singletons/PlayerSingleton.cpp
```diff
@@ -118,3 +163,144 @@ void PlayerSingleton::unset_selected_province() {
...
+	InstanceManager* instance_manager = GameSingleton::get_singleton()->get_instance_manager();
+	ERR_FAIL_NULL(instance_manager);
+
+	instance_manager->queue_game_action(
+		game_action_type_t::GAME_ACTION_CREATE_LEADER,
+		std::pair<uint64_t, bool> { player_country->get_index(), is_general }
+	);
+}
+
+void PlayerSingleton::set_can_use_leader(uint64_t leader_id, bool can_use) const {
```
wvpm, line 262: What is the context here?
When does a leader no longer become usable?
Only when he dies, right?

wvpm (approved):

extension/src/openvic-extension/singletons/PlayerSingleton.cpp
Hop311: This is for when you toggle the checkbox on each leader in the military menu, it marks them as usable/unusable by armies/navies. When unusable, leaders get unassigned from their stacks and can't be manually or automatically assigned back

extension/src/openvic-extension/singletons/PlayerSingleton.cpp
Hop311: IMO both are equally clear, general vs admiral, land vs naval, the only way to make it clearer would be to include the fact that false means admiral in the variable name but that would make it too verbose.

PR #384, opened 2025-02-11
Update to use SIM's new military structs, trade helper functions and country+province indices

Hop311:


extension/src/openvic-extension/singletons/MilitaryMenu.cpp
```diff
@@ -329,7 +324,9 @@ Dictionary MenuSingleton::make_in_progress_unit_dict() const {
 	in_progress_unit_dict[military_info_unit_progress_key] = progress.to_float();
 	in_progress_unit_dict[military_info_unit_icon_key] = unit_type->get_icon();
 	in_progress_unit_dict[military_info_unit_name_key] = Utilities::std_to_godot_string(unit_type->get_identifier());
-	in_progress_unit_dict[military_info_unit_location_key] = Utilities::std_to_godot_string(location->get_identifier());
+	if (location != nullptr) {
```
wvpm, line 327: How can location be null here?
All units are constructed/recruited from a location, right?

wvpm (approved):

extension/src/openvic-extension/singletons/MilitaryMenu.cpp
Hop311: Yeah it shouldn't be null, but better to be safe, this is UI code so performance isn't critical

PR #395, opened 2025-02-22
Allow pie chart tooltip customisation and fix Population Menu over-scaling

Hop311:
Relies on `IndexedMap` fix from [OpenVic-Simulation#339](https://github.com/OpenVicProject/OpenVic-Simulation/pull/339)

Fixes a bug with Population Menu where ideology, issue and vote distributions for all currently selected pops (based on selected states/provinces and pop type filters) had mis-scaling and overflow issues due to each pop's contribution being multiplied by its size despite the pop's ideology/issue/vote distribution already being scaled so that it's total weight equals its size.

### DIFF
--- a/extension/src/openvic-extension/classes/GFXPieChartTexture.hpp
+++ b/extension/src/openvic-extension/classes/GFXPieChartTexture.hpp
@@ -7,13 +7,24 @@
 #include "openvic-extension/utility/Utilities.hpp"
 
 namespace OpenVic {
+	template<typename Container>
+	concept IsPieChartDistribution = (
+			/* ordered_map<T const*, mapped_type>, T derived from HasIdentifierAndColour */
+			utility::specialization_of<Container, tsl::ordered_map>
+			/* IndexedMap<T, mapped_type>, T derived from HasIdentifierAndColour */
+			|| utility::specialization_of<Container, IndexedMap>
+		)
+		&& HasGetIdentifierAndGetColour<std::remove_pointer_t<typename Container::key_type>>
+		&& std::convertible_to<typename Container::mapped_type, float>;
+
 	class GFXPieChartTexture : public godot::ImageTexture {
 		GDCLASS(GFXPieChartTexture, godot::ImageTexture)
 
 	public:
 		using godot_pie_chart_data_t = godot::TypedArray<godot::Dictionary>;
 		struct slice_t {
-			godot::String name;
+			godot::String identifier;
+			godot::String tooltip;
 			godot::Color colour;
 			float weight;
 		};
@@ -25,6 +36,7 @@ namespace OpenVic {
 		godot::Ref<godot::Image> pie_chart_image;
 
 		static godot::StringName const& _slice_identifier_key();
+		static godot::StringName const& _slice_tooltip_key();
 		static godot::StringName const& _slice_colour_key();
 		static godot::StringName const& _slice_weight_key();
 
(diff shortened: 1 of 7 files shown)


extension/src/openvic-extension/classes/GFXPieChartTexture.hpp
```diff
@@ -33,61 +45,64 @@ namespace OpenVic {
...
 
-					if (value != 0.0f) {
-						sorted_dist.emplace_back(key, value);
-					}
+			for (auto [key_ref_or_ptr, non_float_value] : distribution) {
+				key_type const* key_ptr;
+				if constexpr (std::same_as<decltype(key_ptr), decltype(key_ref_or_ptr)>) {
+					key_ptr = key_ref_or_ptr;
+				} else {
+					key_ptr = &key_ref_or_ptr;
```
Nemrav, line 70: Does anything bad happen if this is a nullptr? or does the container template stop nullptrs?

extension/src/openvic-extension/singletons/PopulationMenu.cpp
```diff
@@ -605,6 +605,53 @@ Error MenuSingleton::population_menu_select_sort_key(PopSortKey sort_key) {
...
+	tooltips.reserve(distribution.size());
+
+	String full_tooltip = get_tooltip_separator().trim_suffix("\n");
+
+	float total_weight = 0.0f;
+	for (auto [key, weight] : distribution) {
+		total_weight += static_cast<float>(weight);
+	}
+
+	static const String pie_chart_tooltip_format_key = "%s: " + GUILabel::get_colour_marker() + "Y%s%%" +
```
Nemrav, line 624: I understand this produces "label: ##.#%" but I'm uncertain what the "Y" and "!" do. Just for my sake, could you explain that?

extension/src/openvic-extension/singletons/PopulationMenu.cpp
```diff
@@ -605,6 +605,53 @@ Error MenuSingleton::population_menu_select_sort_key(PopSortKey sort_key) {
...
+		GUILabel::get_colour_marker() + "!";
+
+	for (auto [key_ref_or_ptr, weight] : distribution) {
+		if (weight > 0.0f) {
+			key_type const* key_ptr;
+			if constexpr (std::same_as<decltype(key_ptr), decltype(key_ref_or_ptr)>) {
+				key_ptr = key_ref_or_ptr;
+			} else {
+				key_ptr = &key_ref_or_ptr;
```
Nemrav, line 633: ditto to earlier, could something go wrong with the pointers?

Nemrav (approved):
Looks good in game and code-wise, I just want clarification on a couple things.

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Hop311: `Y` (together with the colour marker preceding it) makes the text yellow, `!` (together with its colour marker) reverts the text to the default colour, in this case white

extension/src/openvic-extension/classes/GFXPieChartTexture.hpp
Hop311: There will never be nullptrs in here, the code for populating these containers operates with `const&` references as inputs, we only use pointers here rather than references because you can't cleanly use references in containers (because you can't reassign them)

extension/src/openvic-extension/singletons/PopulationMenu.cpp
Hop311: Nope, see earlier comment

PR #396, opened 2025-02-22
Add GUIListBox sort_children function and use it in Trade and Military menus

Hop311:


Nemrav (approved):
Everything looks good.

PR #419, opened 2025-03-07
Fix reductive SoundEffect lookup failure

Spartan322:


Nemrav (approved):
Works, and looks good.

PR #453, opened 2025-05-16
Fix GitInfo displays always being incorrect

Spartan322:
- Supersedes #450

Move GitInfo retrieval to GDExtension
Add .gdignore to game/export**


extension/src/openvic-extension/singletons/GitInfo.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <godot_cpp/classes/object.hpp>
+#include <godot_cpp/classes/wrapped.hpp>
+#include <godot_cpp/variant/string_name.hpp>
+
+namespace OpenVic {
+	class GitInfo : public godot::Object {
+		GDCLASS(GitInfo, Object);
```
Hop311, line 9: Not a big deal, but in almost all our uses of `GDCLASS` we haven't added a `;` to the end (exceptions: `CursorSingleton` and `SoundSingleton`, and now `GitInfo`)

extension/src/openvic-extension/singletons/GitInfo.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+	protected:
+		static void _bind_methods();
+
+	public:
+		godot::StringName get_commit_hash() const;
+		uint64_t get_commit_timestamp() const;
+		godot::StringName get_short_hash() const;
+		godot::StringName get_tag() const;
+		godot::StringName get_release_name() const;
```
Hop311, line 26: Could any of these be replaced by `PROPERTY`-generated getters? Or would there be some issue from them being `constexpr` and/or having `const&` return types (apart from `get_commit_timestamp`)?

extension/src/openvic-extension/singletons/GitInfo.cpp
```diff
@@ -0,0 +1,80 @@
...
+	OV_BIND_METHOD(GitInfo::get_release_name);
+
+	ADD_PROPERTY(
+		PropertyInfo(Variant::STRING_NAME, "commit_hash", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NONE), //
+		"", "get_commit_hash"
+	);
+	ADD_PROPERTY(
+		PropertyInfo(Variant::STRING_NAME, "commit_timestamp", PROPERTY_HINT_NONE, "", PROPERTY_USAGE_NONE), //
+		"", "get_commit_timestamp"
+	);
```
Hop311, line 43: Shouldn't this have type `Variant::INT`, or whatever else best reflects `uint64_t`, rather than being `StringName` like the rest? Especially given how `GitInfo.xml` lists `type="int"` for `commit_timestamp`.
And could it use some sort of proprty hint or other extra argument to specify that it's a timestamp? Presumably it's a common int type/usage, unless Godot has its own timestamp type?

extension/src/openvic-extension/singletons/GitInfo.cpp
```diff
@@ -0,0 +1,80 @@
+
+#include "GitInfo.hpp"
+
+#include <godot_cpp/core/error_macros.hpp>
+
+#include "openvic-extension/utility/ClassBindings.hpp"
+
+#include "gen/git_info.gen.hpp"
+#include "godot_cpp/classes/global_constants.hpp"
```
Hop311, line 9: Why not have this above with the other `godot_cpp` include, and in `<>` rather than `""`?

extension/src/openvic-extension/register_types.cpp
```diff
@@ -37,6 +39,7 @@ using namespace godot;
 using namespace OpenVic;
 
 static Checksum* _checksum_singleton = nullptr;
+static GitInfo* _git_info_singleton = nullptr;
```
Hop311, line 42: Could you swap these two lines so that the order of declaration matches the order of initialisation and deletion in the functions below (`GitInfo` before `Checksum`)? Or leave these declarations and swap their corresponding code in the functions below to match the order.

extension/src/openvic-extension/singletons/GitInfo.cpp
```diff
@@ -0,0 +1,80 @@
...
+using namespace OpenVic;
+
+GitInfo::GitInfo() {
+	ERR_FAIL_COND(_singleton != nullptr);
+	_singleton = this;
+
+	_game_commit_hash = GAME_COMMIT_HASH.data();
+	_game_commit_short_hash = _game_commit_hash.substr(0, 7);
+	_game_tag = GAME_TAG.data();
+	_game_release = GAME_RELEASE.data();
```
Hop311, line 21:
```suggestion
	_game_commit_hash = Utilities::std_to_godot_string(GAME_COMMIT_HASH);
	_game_commit_short_hash = _game_commit_hash.substr(0, 7);
	_game_tag = Utilities::std_to_godot_string(GAME_TAG);
	_game_release = Utilities::std_to_godot_string(GAME_RELEASE);
```
So here we initialise `StringName`s from `char const*`s obtained via `std::string_view`s. We know in practice that these C-strings will always be null-terminated (as the `std::string_view`s are intialised from string literals), but would the `std::string_view` actually include the null character in its length?
To be absolutely sure we avoid any such problem in the future, we could use `Utilities::std_to_godot_string` in all 3 cases here rather than just getting `data()` (and add `#include "openvic-extension/utility/Utilities.hpp"` above).

SConstruct
```diff
@@ -32,6 +32,28 @@ env.Prepend(LIBS=godot_env["LIBS"])
...
+#pragma once
+#include <cstdint>
+#include <string_view>
+
+namespace OpenVic {{
+    static constexpr std::string_view GAME_TAG = "{git_tag}";
+    static constexpr std::string_view GAME_RELEASE = "{git_release}";
+    static constexpr std::string_view GAME_COMMIT_HASH = "{git_hash}";
+    static constexpr const uint64_t GAME_COMMIT_TIMESTAMP = {git_timestamp};
+}}
```
Hop311, line 49:
```suggestion
#pragma once

#include <cstdint>
#include <string_view>

namespace OpenVic {{
	static constexpr std::string_view GAME_TAG = "{git_tag}";
	static constexpr std::string_view GAME_RELEASE = "{git_release}";
	static constexpr std::string_view GAME_COMMIT_HASH = "{git_hash}";
	static constexpr const uint64_t GAME_COMMIT_TIMESTAMP = {git_timestamp};
}}
```
Any chance we could use actual tab characters so the generated header better matches our C++ format? I do realise this raises the issue of introducing tab characters to an otherwise 4-space-indented file, so maybe it could use `\t` instead?
Also it would be nice to add a newline after `#pragma once`, again to better match our wider C++ style, but this whole comment is a low priority nit-pick anyway 😅
(Also would it be worth using f-strings and/or r-strings? I'm not sure what's best practice in modern Python)

Hop311 (approved):
Tested on Windows, all working 👍
Left some comments, mostly very minor, lemme know what you think

SConstruct
Spartan322: > Any chance we could use actual tab characters 

Already does

> so maybe it could use \t instead?

No, that makes it less readable

> Also it would be nice to add a newline after #pragma once

There already is one

> (Also would it be worth using f-strings and/or r-strings? I'm not sure what's best practice in modern Python)

Its already using f-strings.

extension/src/openvic-extension/singletons/GitInfo.hpp
Spartan322: Following Godot's format it would be expected to have the semicolon after GDCLASS, generally we should prefer to conform to them as it makes things more consistent and easier imo.

extension/src/openvic-extension/singletons/GitInfo.hpp
Spartan322: For something to be constexpr, its definition needs to be visible to the translation unit (mean the source file being compiled) as its being compiled and not at link-time.

extension/src/openvic-extension/singletons/GitInfo.hpp
Hop311: Ah ok, the `godot_cpp` headers use `GDEXTENSION_CLASS` without a semicolon, but I see in the actual `godot` code `GDCLASS` is indeed followed by a semicolon. TBH I don't mind which we use, just that we apply it consistently across `openvic-extension`.

extension/src/openvic-extension/singletons/GitInfo.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+GitInfo::GitInfo() {
+	ERR_FAIL_COND(_singleton != nullptr);
+	_singleton = this;
+
+	commit_hash = Utilities::std_to_godot_string(GAME_COMMIT_HASH);
+	short_hash = commit_hash.substr(0, 7);
+	tag = Utilities::std_to_godot_string(GAME_TAG);
+	release_name = Utilities::std_to_godot_string(GAME_RELEASE);
+	commit_timestamp = GAME_COMMIT_TIMESTAMP;
```
Hop311, line 23: Very minor comment: the variable order is different here vs elsewhere, specifically `commit_timestamp` is last here vs second in the header and in the `ADD_PROPERTY` list below (although it is also last in the generated file).

SConstruct
```diff
@@ -32,6 +32,29 @@ env.Prepend(LIBS=godot_env["LIBS"])
...
+#pragma once
+
+#include <cstdint>
+#include <string_view>
+
+namespace OpenVic {{
+	static constexpr std::string_view GAME_TAG = "{git_tag}";
+	static constexpr std::string_view GAME_RELEASE = "{git_release}";
+	static constexpr std::string_view GAME_COMMIT_HASH = "{git_hash}";
+	static constexpr const uint64_t GAME_COMMIT_TIMESTAMP = {git_timestamp};
```
Hop311, line 49: It might be worth adding a suffix to the timestamp integer literal to ensure it's interpreted correctly (i.e. unsigned).
When testing I saw big enough integer literal values cause a warning:
```cpp
// Warning: Integer literal is too large to be represented in a signed integer type, interpreting as unsigned
static constexpr const uint64_t N = 18446744073709551615;
```

Hop311 (approved):

SConstruct
Spartan322: Can you tell me what running `git log  --pretty=format:%ct --no-show-signature HEAD` in the sim project root produces on your local machine? Cause that's way too large to be a correct unix timestamp in seconds.

extension/src/openvic-extension/singletons/GitInfo.hpp
Spartan322: I would like to solve that completely in #448

SConstruct
Hop311: Ah yeah it produces values around `1747694360`, the number above was just testing to see if/when using integer literals for unsigned values can have issues

Hop311 (approved):

PR #454, opened 2025-05-18
Remove optional storage of godot::Ref values

Spartan322:
Remove optional storage of godot packed arrays

### DIFF
--- a/extension/src/openvic-extension/singletons/SoundSingleton.cpp
+++ b/extension/src/openvic-extension/singletons/SoundSingleton.cpp
@@ -180,11 +180,12 @@ Ref<AudioStreamWAV> SoundSingleton::get_sound(String const& path) {
 
 // Get a sound by its define name
 Ref<AudioStreamWAV> SoundSingleton::get_sound_stream(String const& path) {
-	if (sfx_define[path].audioStream.has_value()) {
-		return sfx_define[path].audioStream.value();
-	}
+	sfx_define_map_t::iterator it = sfx_define.find(path);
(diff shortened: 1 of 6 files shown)


Spartan322:
Issue is we're already doing option 1 because nullptr is assignable to Ref, so in every case we accessed the object the Ref was already being assigned, you mixed up nullopt and nullptr, the former is assignable to any std::optional, the latter is assignable to Ref and does not constitute a lack of a Ref value.

extension/src/openvic-extension/singletons/SoundSingleton.cpp
```diff
@@ -180,11 +180,12 @@ Ref<AudioStreamWAV> SoundSingleton::get_sound(String const& path) {
 
 // Get a sound by its define name
 Ref<AudioStreamWAV> SoundSingleton::get_sound_stream(String const& path) {
-	if (sfx_define[path].audioStream.has_value()) {
-		return sfx_define[path].audioStream.value();
-	}
+	sfx_define_map_t::iterator it = sfx_define.find(path);
```
Hop311, line 183:
```suggestion
	const sfx_define_map_t::const_iterator it = sfx_define.find(path);
```
Could this be const? (same for any other similar cases across the repo)

Hop311 (approved):

extension/src/openvic-extension/singletons/SoundSingleton.cpp
Spartan322: It could be but it wouldn't do anything to do so, its just excessively verbose.

PR #507, opened 2025-07-07
Refactor scrollbar

wvpm:
The scrollbar was bugged and difficult to work with.
Having to specify a scale when adjusting range limits is redundant.
Min max values and step size also are redundant.

All we need is a scrollbar that starts at 0 and increments by 1 to a max (step_count) combined with a fixed_point_t offset and scale.
I implemented the scale using numerator/denominator as neither floats nor fixed_point_t can store 1/100 accurately.

Also setting range limits on scrollbars that are not range limited no longer results in errors if the range limit are 0 to step_count. This allows us to set from slider value.

Per @Spartan322 's suggestion `godot::Error` returns are minimised and method calls with invalid data are merely logged and ignored. Also signals are blocked instead of passing a bool to indicate whether to signal or not. Blocking signals now also blocks c++ signal.

The trade menu's stockpile slider logic is optimised and the special case of min > max from [SliderValue.hpp](https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/types/SliderValue.hpp) is now supported in the UI.

### DIFF
--- a/extension/src/openvic-extension/classes/GUIListBox.cpp
+++ b/extension/src/openvic-extension/classes/GUIListBox.cpp
@@ -63,7 +63,11 @@ Error GUIListBox::_calculate_max_scroll_index(bool signal) {
 
 	ERR_FAIL_NULL_V(scrollbar, FAILED);
 
-	scrollbar->set_limits(0, max_scroll_index, false);
+	const bool was_blocking_signals = scrollbar->is_blocking_signals();
+	scrollbar->set_block_signals(true);
+	scrollbar->set_step_count(max_scroll_index);
+	scrollbar->set_block_signals(was_blocking_signals);
+
 	scrollbar->set_visible(max_scroll_index > 0);
 
 	set_scroll_index(scrollbar->get_value(), signal);
@@ -209,7 +213,10 @@ void GUIListBox::set_scroll_index(int32_t new_scroll_index, bool signal) {
 	scroll_index = std::clamp(new_scroll_index, 0, max_scroll_index);
 
 	if (scrollbar != nullptr && scrollbar->get_value() != scroll_index) {
-		scrollbar->set_value(scroll_index, false);
+		const bool was_blocking_signals = scrollbar->is_blocking_signals();
+		scrollbar->set_block_signals(true);
+		scrollbar->set_value(scroll_index);
+		scrollbar->set_block_signals(was_blocking_signals);
 	}
 
 	if (signal && scroll_index != old_scroll_index) {
--- a/extension/src/openvic-extension/classes/GUIScrollbar.cpp
+++ b/extension/src/openvic-extension/classes/GUIScrollbar.cpp
@@ -1,10 +1,15 @@
 #include "GUIScrollbar.hpp"
 
+#include <cstdint>
+
+#include <godot_cpp/classes/global_constants.hpp>
 #include <godot_cpp/classes/input_event_mouse_button.hpp>
 #include <godot_cpp/classes/input_event_mouse_motion.hpp>
 #include <godot_cpp/variant/utility_functions.hpp>
 
+#include <openvic-simulation/types/fixed_point/FixedPoint.hpp>
 #include <openvic-simulation/types/SliderValue.hpp>
+#include <openvic-simulation/utility/Logger.hpp>
 
 #include "openvic-extension/utility/ClassBindings.hpp"
 #include "openvic-extension/utility/UITools.hpp"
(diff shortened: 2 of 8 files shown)


extension/src/openvic-extension/classes/GUIScrollbar.cpp
```diff
@@ -1,9 +1,13 @@
 #include "GUIScrollbar.hpp"
 
+#include <cstdint>
+
+#include "godot_cpp/classes/global_constants.hpp"
```
Spartan322, line 5:
```suggestion
#include <godot_cpp/classes/global_constants.hpp>
```

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -168,9 +168,12 @@ Dictionary MenuSingleton::get_trade_menu_trade_details_info(
 	if (stockpile_cutoff_slider != nullptr) {
 		int32_t index = 0;
 
-		while (index < stockpile_cutoff_slider->get_max_value() && calculate_trade_menu_stockpile_cutoff_amount_fp(
-			index * stockpile_cutoff_slider->get_step_size()
-		) < good_data.stockpile_cutoff) {
+		while (
+			index < stockpile_cutoff_slider->get_step_count()
+			&& calculate_trade_menu_stockpile_cutoff_amount_fp(
```
wvpm, line 173: Still causes issues. I'm working on it.
I could get the ratio of the slider and multiply that by 2000, but that would diverge from Victoria 2 where the gui file defines the max value.

extension/src/openvic-extension/classes/GUIScrollbar.hpp
```diff
@@ -1,13 +1,15 @@
 #pragma once
 
+#include <cstdint>
 #include <godot_cpp/classes/control.hpp>
```
Spartan322, line 4:
```suggestion
#include <cstdint>

#include <godot_cpp/classes/control.hpp>
```

extension/src/openvic-extension/classes/GUIScrollbar.cpp
```diff
@@ -489,83 +503,109 @@ float GUIScrollbar::get_value_as_ratio() const {
...
+		new_upper_range_limit.has_value()
+			? _fp_to_value(new_upper_range_limit.value())
+			: std::optional<fixed_point_t>{},
 		signal
 	);
 }
-
 Error GUIScrollbar::set_range_limits_and_value_from_slider_value(
-	SliderValue const& slider_value, int32_t scale, bool signal
+	SliderValue const& slider_value, bool signal
```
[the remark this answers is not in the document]
wvpm: Not good enough as it also emits c++ signals.
I'd prefer not to emit signals when programettically changing values so I pass false basically everywhere.
The option existed to pass true in the old version, so I kept that to prevent further breaking the interface.

extension/src/openvic-extension/singletons/TradeMenu.cpp
wvpm: Fixed it very nicely :)

extension/src/openvic-extension/classes/GUIListBox.cpp
```diff
@@ -209,7 +213,9 @@ void GUIListBox::set_scroll_index(int32_t new_scroll_index, bool signal) {
 	scroll_index = std::clamp(new_scroll_index, 0, max_scroll_index);
 
 	if (scrollbar != nullptr && scrollbar->get_value() != scroll_index) {
-		scrollbar->set_value(scroll_index, false);
+		scrollbar->set_block_signals(true);
+		scrollbar->set_value(scroll_index);
+		scrollbar->set_block_signals(false);
```
Spartan322, line 218: Shouldn't we restore the block signal value instead?

extension/src/openvic-extension/singletons/TradeMenu.cpp
```diff
@@ -166,17 +166,15 @@ Dictionary MenuSingleton::get_trade_menu_trade_details_info(
...
-
-		while (index < stockpile_cutoff_slider->get_max_value() && calculate_trade_menu_stockpile_cutoff_amount_fp(
-			index * stockpile_cutoff_slider->get_step_size()
-		) < good_data.stockpile_cutoff) {
-			++index;
-		}
-
-		// TODO - use a more efficient algorithm, e.g. some kind of binary search
+		const int32_t index = calculate_slider_value_from_trade_menu_stockpile_cutoff(
+			1984,
```
wvpm, line 170: Undo test code

PR #509, opened 2025-07-08
Refactor scrollbar

Spartan322:
- Supersedes #507 

The scrollbar was bugged and difficult to work with.
Having to specify a scale when adjusting range limits is redundant.
Min max values and step size also are redundant.

All we need is a scrollbar that starts at 0 and increments by 1 to a max (step_count) combined with a fixed_point_t offset and scale.
I implemented the scale using numerator/denominator as neither floats nor fixed_point_t can store 1/100 accurately.

Also setting range limits on scrollbars that are not range limited no longer results in errors if the range limit are 0 to step_count. This allows us to set from slider value.

Per @Spartan322 's suggestion `godot::Error` returns are minimised and method calls with invalid data are merely logged and ignored. Also signals are blocked instead of passing a bool to indicate whether to signal or not. Blocking signals now also blocks c++ signal.

The trade menu's stockpile slider logic is optimised and the special case of min > max from [SliderValue.hpp](https://github.com/OpenVicProject/OpenVic-Simulation/blob/master/src/openvic-simulation/types/SliderValue.hpp) is now supported in the UI.

### DIFF
--- a/extension/doc_classes/GUIScrollbar.xml
+++ b/extension/doc_classes/GUIScrollbar.xml
@@ -19,7 +19,6 @@
 		</method>
 		<method name="decrement_value">
 			<return type="void" />
-			<param index="0" name="signal" type="bool" default="true" />
 			<description>
 			</description>
 		</method>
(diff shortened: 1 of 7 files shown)


game/src/UI/Session/BudgetMenu.gd
```diff
@@ -236,7 +236,6 @@ func _update_info() -> void:
 			_balance_label.text = "§%s%s¤" % ["G+" if balance > 0.0 else "R" if balance < 0.0 else "Y+", GUINode.float_to_string_dp_dynamic(balance)]
 
 		# TODO - set strata tax and debt charts
-		# TODO - update sliders to reflect changes in limits
```
wvpm, line 239: Part of #510

wvpm (approved):

PR #527, opened 2025-07-22
Optimise godot::vformat by pooling args array

wvpm:
the godot engine has a cpp implementation that emulates the godot script features. One commonly used feature is string formatting. In godot script you can just use `%`. In godot-cpp you use `godot::vformat(template, ...)`.

The issue is this is implemented as follows:
```cpp
template <typename... VarArgs>
String vformat(const String &p_text, const VarArgs... p_args) {
    Variant args[sizeof...(p_args) + 1] = { p_args..., Variant() }; // +1 makes sure zero sized arrays are also supported.
    Array args_array;
    args_array.resize(sizeof...(p_args));
    for (uint32_t i = 0; i < sizeof...(p_args); i++) {
        args_array[i] = args[i];
    }

    return p_text % args_array;
}
```

So it creates a new array each invocation.
The array is always discarded afterwards, resulting in redundant allocations and deallocations.

Instead we should clear and reuse the array. To support multithreading, the reusable array must be thread local.
Since nesting could occur, we use a vector of reusable arrays.

Note that clearing the array would free the memory and resizing the array would reallocate memory. So resize is called once to set the array to the desired size and clear is avoided. Instead the array is filled with empty values. This is done to prevent references being kept alive by simply being in the pooled array.

### DIFF
--- a/extension/src/openvic-extension/utility/Utilities.hpp
+++ b/extension/src/openvic-extension/utility/Utilities.hpp
@@ -5,6 +5,7 @@
 #include <godot_cpp/classes/file_access.hpp>
 #include <godot_cpp/classes/font_file.hpp>
 #include <godot_cpp/classes/image_texture.hpp>
+#include <godot_cpp/variant/variant.hpp>
 
 #include <openvic-simulation/types/Colour.hpp>
 #include <openvic-simulation/types/Date.hpp>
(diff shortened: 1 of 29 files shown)


extension/src/openvic-extension/utility/Utilities.hpp
```diff
@@ -92,6 +92,28 @@ namespace OpenVic::Utilities {
...
+			: std::move(_formatting_array_pool.back());
+		
+		if (!was_empty) {
+			_formatting_array_pool.pop_back();
+		}
+
+    	args_array.resize(sizeof...(p_args));
+    	size_t i = 0;
+    	// Assign arguments directly using a fold expression
+    	([&]() { args_array[i++] = p_args; }(), ...);
```
Spartan322, line 110:
```suggestion
		args_array.resize(sizeof...(p_args));
		size_t i = 0;
		// Assign arguments directly using a fold expression
		([&]() { args_array[i++] = p_args; }(), ...);
```

extension/src/openvic-extension/utility/Utilities.hpp
```diff
@@ -92,6 +92,28 @@ namespace OpenVic::Utilities {
 		godot::Image::Format format = godot::Image::Format::FORMAT_RGBA8
 	);
 
+	static thread_local memory::vector<godot::Array> _formatting_array_pool;
```
Spartan322, line 95: thread_local is by default static, this aside `_formatting_array_pool` should be hidden

Spartan322 (dismissed):

extension/src/openvic-extension/utility/Utilities.hpp
```diff
@@ -9,6 +9,7 @@
 #include <openvic-simulation/types/Colour.hpp>
 #include <openvic-simulation/types/Date.hpp>
 #include <openvic-simulation/types/Vector.hpp>
+#include "godot_cpp/variant/variant.hpp"
```
wvpm, line 12: wrong import

PR #549, opened 2025-08-15
Top bar refactored to cpp

wvpm:
Refactor of topbar.gd + dictionary from menu singleton to smaller cpp components getting their own data.


extension/src/openvic-extension/components/overview/BudgetOverview.cpp
```diff
@@ -0,0 +1,88 @@
...
+		for (fixed_point_t const& balance : balance_history) {
+			if (balance > maximum_balance) {
+				maximum_balance = balance;
+			}
+			if (balance < minimum_balance) {
+				minimum_balance = balance;
+			}
+		}
+		last_balance = balance_history.back();
+		godot::PackedFloat32Array line_values;
```
wvpm: This should be refactored to directly use the balance_history or its iterators instead of dragging godot into it (`set_gradient_line` merely iterates through the array).

PR #564, opened 2025-08-25
Add province panel list pooling to PopulationMenu

Spartan322:
This fixes the performance obliteration that the province list updates were causing because they were constantly freeing and then allocating panels instead of reusing them. Provinces are now pooled by type and added and removed from the SceneTree instead.

### DIFF
--- a/game/src/UI/Session/PopulationMenu.gd
+++ b/game/src/UI/Session/PopulationMenu.gd
@@ -8,6 +8,12 @@ const _scene_name : String = "country_pops"
 
 var _pop_screen_panel : Panel
 
+var _province_list_type_pool : Dictionary = {
+	MenuSingleton.LIST_ENTRY_COUNTRY: [],
+	MenuSingleton.LIST_ENTRY_STATE: [],
+	MenuSingleton.LIST_ENTRY_PROVINCE: []
+}
+
 var _province_listbox : GUIListBox
 var _province_list_scroll_index : int = 0
 var _province_list_types : Array[MenuSingleton.ProvinceListEntry]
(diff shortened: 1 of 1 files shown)


game/src/UI/Session/PopulationMenu.gd
```diff
@@ -83,38 +89,34 @@ func _ready() -> void:
...
-		_province_list_name_labels.push_back(null)
-		_province_list_size_labels.push_back(null)
-		_province_list_growth_icons.push_back(null)
-		_province_list_colony_buttons.push_back(null)
-		_province_list_national_focus_buttons.push_back(null)
-		_province_list_expand_buttons.push_back(null)
-
-	if _province_list_types[index] == type:
+	if _province_list_types.size() <= index:
+		_resize_province_list(index + 1)
```
wvpm, line 112: Are you resizing often here?
If so, consider simply doubling the size whenever the array is too small.
This takes up a bit more memory but saves resizing for every call.

game/src/UI/Session/PopulationMenu.gd
```diff
@@ -123,14 +125,19 @@ func _generate_province_list_row(index : int, type : MenuSingleton.ProvinceListE
...
-
+	var entry_panel : Panel = _province_list_type_pool[type][index]
 	if not entry_panel:
-		return FAILED
+		entry_panel = GUINode.generate_gui_element(_scene_name, gui_element_names[type])
+		if entry_panel:
+			_province_list_type_pool[type][index] = entry_panel
 
-	_province_list_types[index] = type
+	if entry_panel:
```
wvpm, line 134: Remove check

wvpm (changes requested):
The if `entry_panel` needs to be simplified (see comments).
Besides that looks good

game/src/UI/Session/PopulationMenu.gd
Spartan322: `resize` in godot already handles the growth factor for us, cowdata arrays (which array is) have a growth factor that's always a power of two, so the capacity is always a power of two.

wvpm (approved):

PR #550, opened 2025-08-20
Move ArgumentParser to C++ extension

Spartan322:
Move ArgumentOption to C++ extension
Add help message ANSI terminal coloring

PR #565, opened 2025-08-27
Add GameSettings to replace Events.Options

Spartan322:
- Depends on #550

Remove OptionsMenu hacks to load settings
Separate OpenVic game settings from dataloading settings
Remove SoundTab script

I found it necessary that while trying to cheaply expose the Simulation data to GDScript through a server design, the servers needed a capacity in C++ to access the path settings which without this is non-trivial. This otherwise overhauls the settings so that the startup hacks are unnecessary.


game/src/UI/GameMenu/OptionMenu/OptionsMenu.gd
```diff
@@ -7,6 +7,9 @@ signal back_button_pressed
 
 @export var _tab_container : TabContainer
 
+@onready
+var _settings := GameSettings.load_from_file("user://settings.cfg")
```
wvpm, line 11: "user://settings.cfg"
Should probably be a constant field in GameSettings or encapsulated entirely.
Different menus all using the same magic string is bad.

extension/doc_classes/GameSettings.xml
```diff
@@ -0,0 +1,124 @@
...
+		<method name="load">
+			<return type="int" enum="Error" />
+			<param index="0" name="path" type="String" default="&quot;&quot;" />
+			<description>
+				Loads the settings file specified by [param path] and then emitting [signal Resource.changed]. If [param path] is empty then it is set to [member Resource.resource_path]. The file's contents are parsed and loaded in the [GameSettings] object which the method was called on. Sets [member Resource.resource_path] to [param path].
+				Returns [constant OK] on success, or one of the other [enum Error] values if the operation failed.
+				See [method ConfigFile.load].
+			</description>
+		</method>
+		<method name="load_deprecated_file">
```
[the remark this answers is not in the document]
Spartan322: Because its loading the old file into the new place. This is for when settings change their place which may happen in the future for other things.

game/src/UI/GameMenu/OptionMenu/OptionsMenu.gd
Spartan322: No it shouldn't, this should neither be encapsulated nor stored in GameSettings, GameSettings must remain completely independent of any script functionality, its neither a magic string, there is nowhere to place it, and each object should store its own reference that may change in the future.

wvpm (approved):

PR #613, opened 2025-11-06
Fix some orphan StringNames caused by extension

Spartan322:


Spartan322:
It's probably more valuable to work on marshaling sim data into gdscript and eliminating the dictionary hardcoding before implementing this, 90% of the work behind this PR would disappear in that case.

PR #673, opened 2026-05-20
Remove viewport focus check from MapView _process

Catylist0:
## Summary

- Remove the per-frame viewport focus / `is_input_handled()` check from `MapView._process` that was clearing province hover.
- Track mouse presence over the map via `NOTIFICATION_WM_MOUSE_ENTER` and `NOTIFICATION_WM_MOUSE_EXIT` instead.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/game/src/Systems/Session/Map/MapView.gd
+++ b/game/src/Systems/Session/Map/MapView.gd
@@ -150,9 +150,13 @@ func _ready() -> void:
 
 
 func _notification(what: int) -> void:
-	if what == NOTIFICATION_WM_MOUSE_EXIT:
-		_mouse_over_viewport = false
-		province_unhovered.emit()
+	match what:
+		NOTIFICATION_WM_MOUSE_ENTER:
+			_mouse_over_viewport = true
+			queue_province_hover_update()
+		NOTIFICATION_WM_MOUSE_EXIT, NOTIFICATION_WM_WINDOW_FOCUS_OUT:
+			_mouse_over_viewport = false
+			province_unhovered.emit()
 
 
 func _world_to_map_coords(pos: Vector3) -> Vector2:
@@ -261,7 +265,6 @@ var _cardinal_movement_vector := Vector2.ZERO
 
 func _unhandled_input(event: InputEvent) -> void:
 	if event is InputEventMouseMotion:
-		_mouse_over_viewport = true
 		queue_province_hover_update()
 
 	elif event.is_action(_action_north) or event.is_action(_action_south)\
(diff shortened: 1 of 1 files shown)


game/src/Systems/Session/Map/MapView.gd
```diff
@@ -278,10 +281,6 @@ func _process(delta : float) -> void:
 	if _cardinal_movement_vector != Vector2.ZERO and get_window().gui_get_focus_owner() != null:
 		_cardinal_movement_vector = Vector2.ZERO
 
-	if _is_viewport_inactive():
```
wvpm, line 343: Is this case handled via the `NOTIFICATION_WM_MOUSE_EXIT`?
The old code implies it wasn't.

Catylist0: It should be now, in the old code we weren't catching the on exit stuff iirc, it was just doing a check each tick. Not sure why it was like that, this system should be more correct(?)

wvpm:
From Devin:
_Missing window-focus-loss handling causes indefinite edge scrolling when alt-tabbing

The old code checked _is_viewport_inactive() every frame in _process, which returned true when not get_window().has_focus(). This set _mouse_over_viewport = false whenever the game window lost focus (e.g., alt-tab). The new code replaces this per-frame check with NOTIFICATION_WM_MOUSE_ENTER/NOTIFICATION_WM_MOUSE_EXIT notifications, but these only fire when the mouse cursor physically enters/leaves the window rect — they do not fire when the window merely loses focus while the cursor remains inside it.

Consequently, if the user alt-tabs away while the mouse cursor is near a window edge, _mouse_over_viewport stays true and _mouse_pos_viewport freezes at the last known position. Since _process continues running, _edge_scrolling_vector() (MapView.gd:316-323) returns a non-zero vector every frame, causing the camera to scroll indefinitely until the user returns to the game._

To be fair, that is what Paradox games also do, so I suppose you nailed it.

game/src/Systems/Session/Map/MapView.gd
```diff
@@ -131,9 +131,13 @@ func _ready() -> void:
...
 
 func _notification(what: int) -> void:
-	if what == NOTIFICATION_WM_MOUSE_EXIT:
-		_mouse_over_viewport = false
-		province_unhovered.emit()
+	match what:
+		NOTIFICATION_WM_MOUSE_ENTER:
+			_mouse_over_viewport = true
+			queue_province_hover_update()
+		NOTIFICATION_WM_MOUSE_EXIT:
```
Spartan322:
```suggestion
		NOTIFICATION_WM_MOUSE_EXIT:
		NOTIFICATION_WM_WINDOW_FOCUS_OUT:
```

BrickPi (approved):

PR #681, opened 2026-06-04
Fix Victoria 2 folder selection dialog

Spartan322:
Add repeated select attempts if the selection is wrong (cancel/close button triggers game close)
Fix repeated failure alert popup
Fix folder selection failing on selecting but not opening folder
Fix loading thread running despite empty base path
Fix dialog ok button's lack of translation


Catylist0 (approved):
Functions as expected

PR #700, opened 2026-06-27
Add gettext template generation

Spartan322:
Replace translation tag strings with (British) English strings
Fix erroneous monitor selection setting behavior
Add openvic plugin
Update gdstyle pre-commit hook to v0.2.4

Note: Ignores multiplayer facade

PR #732, opened 2026-07-18
Update pre-commit hooks

Spartan322:
Update ruff-check/ruff-format to v0.15.22
Update mypy to v2.3.0
Update typos to v1.48.0
Update clang-tidy dependency to 22.1.8
Apply clang-tidy fixes

- Leaving the gdstyle hook up to #700


BrickPi (dismissed):
Would be ideal if we could get dependabot to check for and open PRs for pre-commit deps.

BrickPi (approved):

PR #772, opened 2026-08-21
Fix submodule modification CI build skip bug

Spartan322:
In a case where only the submodules in openvic-simulation is changed, changed-files may skip CI build checks, this fixes that bug.


BrickPi (approved):

Spartan322:
Turns out this does not fix the issue where submodules inside a submodule are accounted for, need to go back to the drawing board.

Spartan322:
See tj-actions/changed-files#2140

